PR #143, opened 2022-05-04
chore(build): add more TS Types

ghiscoding:


z3dev:
I think you will be far better off by adding a small eslint configuration to the package.json. Further parameters are then possible as well.

ghiscoding:
thanks but I already have an [.eslintrc](https://github.com/ghiscoding/lerna-lite/blob/main/.eslintrc), the commit that you might have noticed was to fix this warnings which was always there but it was just annoying to see every time :D

```sh
> lint
> eslint packages/** --ext .ts


/home/runner/work/lerna-lite/lerna-lite/packages/tsconfig.bundle.json
  0:0  warning  File ignored because of a matching ignore pattern. Use "--no-ignore" to override

/home/runner/work/lerna-lite/lerna-lite/packages/tsconfig.json
  0:0  warning  File ignored because of a matching ignore pattern. Use "--no-ignore" to override

✖ 2 problems (0 errors, 2 warnings)
```
with the commit change it no longer show, finally, so that SO answer was correct

z3dev:
We got burned by the same glob mistake. It's pretty easy to control but you have to remember that those ** get translated by the shell. Something like this...

`   "lint": "standardx 'packages/**/*.js'",`

PR #168, opened 2022-05-16
feat(version): support for updating pnpm-lock.yaml

JacobSoderblom:
Had problems with the editorconfig in my editor, that's why I added prettier.

Please come with some feedback, I'm really in need of this to work.


JacobSoderblom:
> hmm yeah I didn't have prettier setup but I see that Lerna itself has it, so we can maybe keep it but it shows a lot of changes which is the not fun part. Can you however add more context? What was your issue, what does this PR fixes?

Yea sure!

I'm using pnpm as package manager in a monorepo and using your package to create new versions and such. Everything works like a charm, except when a new version of `package-a` gets applied to `package-b` due to that the pnpm lockfile does not get updated. So in my CI pipeline fails due to there are changes to the lockfile after a new version created lerna-lite.

I hope that gives you some more information.

.prettierrc
```diff
@@ -0,0 +1,7 @@
+{
+  "useTabs": false,
+  "tabWidth": 2,
+  "printWidth": 100,
+  "singleQuote": true,
+  "trailingComma": "none"
+}
```
ghiscoding, line 7: if you're adding Prettier, then it's probably better to also add the VSCode setting for this, I don't want Prettier conflict with my other projects but I do want it enable for this project. I have it setup in a work project with this

`.vscode > settings.json`
```json
{
  "editor.defaultFormatter": "esbenp.prettier-vscode",
  "editor.formatOnSave": true,
  "editor.formatOnPaste": false
}
```

packages/version/src/__tests__/update-lockfile-version.spec.js
```diff
@@ -72,7 +78,36 @@ describe('npm modern lock file', () => {
...
+      '/'
+    ]);
     expect(fs.readJSONSync(rootLockFilePath)).toMatchSnapshot();
   });
-});
\ No newline at end of file
+});
+
+describe('pnpm lock file', () => {
+  test('updateModernLockfileVersion v2 on pnpm lock file', async () => {
```
ghiscoding, line 91: I don't think "updateModernLockfileVersion v2" has anything to do with pnpm. I added this "modern" name mainly for npm lock file v2 (npm >=7) which is quite different from npm <=6). So anyway, since you're adding pnpm, we probably don't need "modern" and "v2" in this test description, it's probably better to rename this.

packages/version/src/version-command.ts
```diff
@@ -80,9 +81,10 @@ export class VersionCommand extends Command<VersionCommandOption> {
   }
 
   get requiresGit(): boolean {
-    return (
-      this.commitAndTag || this.pushToRemote || this.options.allowBranch || this.options.conventionalCommits
-    ) as boolean;
+    return (this.commitAndTag ||
+      this.pushToRemote ||
+      this.options.allowBranch ||
+      this.options.conventionalCommits) as boolean;
```
ghiscoding, line 87: I don't link how Prettier reformatted this line, can you change it to this if possible
```ts
return (
  this.commitAndTag ||
  this.pushToRemote ||
  this.options.allowBranch ||
  this.options.conventionalCommits
) as boolean;
```

JacobSoderblom:
> > Yea sure!
> > I'm using pnpm as package manager in a monorepo and using your package to create new versions and such. Everything works like a charm, except when a new version of `package-a` gets applied to `package-b` due to that the pnpm lockfile does not get updated. So in my CI pipeline fails due to there are changes to the lockfile after a new version created lerna-lite.
> > I hope that gives you some more information.
> 
> Yes that helps in confirming what the PR is for, I didn't think about the yarn/pnpm lock files when adding this `workspace:` protocol 😄
> 
> Even though I worked on adding workspace protocol, I'm not yet using it but I'm looking to change my own monorepo to pnpm as well in the next couple weeks. So any fixes like this are very well appreciated 😉
> 
> I would just like to confirm another thing though, have you tested the code change with a real release or just unit tests? If you did not, then I would suggest to try running the version/publish with the `--git-dry-run` flag which I added for that reason, it won't do any git commits but it should update the lock file that you can inspect and make sure it is in the format you expect (that is how I tested the modern npm lock file code change I did). That might be a bit more tricky to test with pmpn though, perhaps pnpm link Lerna-Lite code!? Anyway, I just want to make sure you tested this properly before merging this. I can try the PR tonight to make sure that it doesn't conflict with npm itself (which this project is based on and just running with `--git-dry-run` within Lerna-Lite code should be sufficient). (update) I tested with Lerna-Lite itself and yes the npm lock file got updated as intended, so that test is good on my side.
> 
> Also another question, you titled this as a `(feat)` but wouldn't that be more of a fix? The reason I'm asking is because this would automatically bump to a minor as `1.4.0`. is that really what you want? Maybe another question, I'm just curious, are you using this as well with the `workspace:` protocol that I shipped recently, is that going to help with that?

Thanks for the feedback, I'll fix all the things you pointed out and the commit msg :)

packages/version/src/lib/update-lockfile-version.ts
```diff
@@ -1,32 +1,97 @@
...
+  return {
+    json,
+    version,
+    path: lockfilePath,
+    packageManager: 'pnpm',
+  };
+}
+
+export async function loadLockfile(cwd: string) {
+  const pm = await preferredPM(cwd);
```
ghiscoding, line 79: is that new dependency really necessary? We already have the `npmClient` command option for this purpose and I would assume that to be sufficient. So I'm not too keen on adding a new lib (of 5kb) when `npmClient` is probably enough, or the original code to just trying loading the file which is what you had in your original PR.

JacobSoderblom: What that package does is checking which package manager you are using, I cannot find any `npmClient` that you are talking about?

JacobSoderblom: Yea sure, I can rebuild that!

ghiscoding:
@JacobSoderblom do you need any help, or time, to advance this PR? I'm migrating my monorepo to pnpm and your PR is probably the last thing needed. Let me know if you need anything, thanks

ghiscoding:
@JacobSoderblom do you need any help, or time, to advance this PR? I'm migrating my monorepo to pnpm and your PR is probably the last thing needed. Let me know if you need anything, thanks

packages/version/src/lib/update-lockfile-version.ts
ghiscoding: hey I looked at the code of `preferred-pm` and it's not 5kb (not sure why it shows 5kb on npm), however it uses [which-pm](https://github.com/zkochan/packages/tree/master/which-pm) created by the same person and I would suggest you to use that one instead, it's quite small and it has less dependencies compare to `preferred-pm` and I assuming it would do what you want to achieve just the same and I agree that it would be easier to use `which-pm` compare to us trying to read every possible lock file.

packages/version/src/__tests__/__snapshots__/update-lockfile-version.spec.js.snap
```diff
@@ -11,7 +11,7 @@ Object {
     },
     "@my-workspace/package-2": Object {
       "requires": Object {
-        "@my-workspace/package-1": "^2.4.0",
+        "@my-workspace/package-1": "^2.3.4",
```
ghiscoding, line 14: this, and all other changes in the snapshot, should really stay at `2.4.0`, so there must be something wrong with your new code

JacobSoderblom:
> @JacobSoderblom do you need any help, or time, to advance this PR? I'm migrating my monorepo to pnpm and your PR is probably the last thing needed. Let me know if you need anything, thanks

Hi!

Sorry for not being active, have not been able to put any time into this for some days then it was holidays in my country :)

I can see that you have done some fixes! That's great! :)

ghiscoding:
@JacobSoderblom 
no worries, I did put in some fixes but I can't merge this as it is since I found out that it doesn't cover every use case
- when a user changes from let say "^1.3.0" to "workspace:^" then current code will not detect the change and will apply "workspace:^1.3.0" instead of "workspace:^"
- when a dependency has different version number or different workspace protocol, the current code will apply the same version regardless of what was there prior. So let say that `package-a` is a dependency of both `package-b` (with "workspace:^1.3.0") and `package-c` (with "1.3.0"), it will apply (assuming a minor bump) "workspace:^1.4.0" everywhere and so `package-c` will lose the fixed version. 
- another thing I saw is that the write yaml removes all extra empty lines that pnpm added and by doing so changes the yaml file structure a lot. It's probably not a huge deal but I'm sure some users will notice it. 

Maybe you can look at the last point for the file structure, but for the other 2 points, I started experimenting by moving the code of where it apply the changes to do it directly at the source of when it updates all `package.json`. in other words, why are we dealing with the lock file after all `package.json` of each packages are already updated? why can't we do it at the same time, so that this way we will send exactly what is sent to the `package.json` and the lock file and they will always match, it also removes all the logic required in each lock file, we can simply use the same version that is sent to the `package.json`, no logic required we just need to find the same list of dependencies in the lock file and update them. So this approach would be to use `package.ts` file when it updates the `package.json` in the following method and that is what I'm experimenting

https://github.com/ghiscoding/lerna-lite/blob/9cd6a13478d02c9028e1ddb6a7c7346fa7221e99/packages/core/src/package.ts#L250-L259

It will probably take me over a week to get it all done, so I will create a new release today or tomorrow but without this PR because I don't think it's a good idea to merge as is. Again if you look at the last point, it would be nice and I'm experimenting with first 2 points

ghiscoding:
@JacobSoderblom 
Could you please test with a new approach that I experimented with in PR #199, this looks much easier and more stable since we take advantage of the pnpm client directly to update the lock file which is much safer and faster.

I do however want to keep your contributions, this is important to me, and I will refactor this PR soon to have at least the Prettier install you brought and also some small changes done in the update lock file. That way I get to keep your valuable contribution. 

Thank you

ghiscoding:
@JacobSoderblom 
So I hope you don't mind 🙏🏻 but after merging PR #199, which to use the client directly with `pnpm install lockfile-only` to update the pnpm lockfile (other clients were added as well), then there's less of a need for your original PR. However it felt really important for me to keep your contributions and so I decided to transform your original PR into a PR that only adds Prettier (in PR #182 that expanded on your original PR) to the project so that I get to keep some of your contributions and you end up showing as a Contributor to this project and hopefully you'll be happy with this. Thanks a lot for all the time you spent in creating this PR and since I already merged #182, I don't see the need to keep this PR open (your contributions were moved to PR #182 anyway). 

I should be able to release a new version in the coming 1 or 2 days, the new flag that you will want to use in the future is [`--sync-workspace-lock`](https://github.com/ghiscoding/lerna-lite/tree/main/packages/version#--sync-workspace-lock), like I said it's not released yet but since the PR #199 is already merged, you can take a sneak peek at the readme docs 😉 

Again thanks a ton for your contribution 🚀

PR #199, opened 2022-06-03
feat(lock): add version `--sync-workspace-lock` flag to update lock file

ghiscoding:
## Description
This flag would run `npm install --package-lock-only` or associated command depending on the package manager client defined in `npmClient` (npm, pnpm or yarn). It will try to use the package manager client to update the lock file instead of relying on Lerna-Lite to update the file which requires a lot more code.

## Motivation and Context
in summary this simplifies a lot the process of updating the lock file and also make it a lot more reliable since we are relying solely on the client to its job with a lot less logic to implement in Lerna-Lite. Relying on the client is also future proof in case the lock file structure ever change.

## How Has This Been Tested?
unit tests were added

## Types of changes

## Checklist:


packages/cli/src/cli-commands/cli-version-commands.ts
```diff
@@ -151,6 +151,11 @@ exports.builder = (yargs, composed) => {
       hidden: true,
       type: 'boolean',
     },
+    'package-lockfile-only': {
+      describe: 'Runs `npm install --package-lock-only` or associated command depending on the package manager defined in `npmClient`.',
+      type: 'boolean',
+      choices: ['gitlab', 'github'],
```
StarpTech, line 157: github?

packages/cli/src/cli-commands/cli-version-commands.ts
```diff
@@ -180,6 +185,15 @@ exports.builder = (yargs, composed) => {
       requiresArg: true,
       defaultDescription: 'v',
     },
+    'no-update-root-lock-file': {
```
StarpTech, line 188: Is this the way how negations are handled?

packages/version/README.md
```diff
@@ -444,6 +445,31 @@ lerna version prepatch --preid next
 When run with this flag, `lerna version` will increment `premajor`, `preminor`, `prepatch`, or `prerelease` semver
 bumps using the specified [prerelease identifier](http://semver.org/#spec-item-9).
 
+### `--package-lockfile-only`
```
StarpTech, line 448: I'd also mention that the workspace: protocol is recommended when using pnpm because it makes it impossible to fetch packages from the registry accidentally.

packages/cli/src/cli-commands/cli-version-commands.ts
ghiscoding: oops, copy + paste thingy... :D

packages/cli/src/cli-commands/cli-version-commands.ts
ghiscoding: yeah that is how yargs, which is what Lerna uses for the CLI, does

ghiscoding:
> > @ghiscoding I have a repro where NPM works fine like pnpm. Again as long as the version constraint match with the local packages, no error is thrown [test-npm.zip](https://github.com/ghiscoding/lerna-lite/files/8831019/test-npm.zip). Could you share a reproducible example with your error?
> 
> And according to [npm/cli#3637](https://github.com/npm/cli/issues/3637) only npm@>=8.5.0 implemented it properly.

@StarpTech 
Thanks a lot for the comment you wrote in the other issue, this was a really great finding and now my code supports both older/newer npm version which is awesome. Here's the [commit](https://github.com/ghiscoding/lerna-lite/pull/199/commits/1207041ebdc41bc0c69a00b7d78602ea604fbe33) I managed to work on over long lunch :D  

All your feedback is very valuable, learning a lot of things 😉

lerna.json
```diff
@@ -12,6 +12,7 @@
       "createRelease": "github",
       "gitDryRun": false,
       "noPrivate": true,
+      "packageLockfileOnly": true,
```
StarpTech, line 15: I'd assume this flag is enabled by default when the user uses a workspace feature from the package manager.

packages/core/src/models/command-options.ts
```diff
@@ -259,6 +262,12 @@ export interface VersionCommandOption {
   /** Defaults to 'v', customize the tag prefix. To remove entirely, pass an empty string. */
   tagVersionPrefix?: string;
 
+  /** Do not update the project root lock file. */
+  noUpdateRootLockFile?: boolean;
```
StarpTech, line 266: Whats the difference between `updateRootLockFile` and `packageLockfileOnly` ?

StarpTech (changes requested):

packages/version/src/version-command.ts
```diff
@@ -658,15 +659,27 @@ export class VersionCommand extends Command<VersionCommandOption> {
...
+            if (semver.gte(localNpmVersion, '8.5.0')) {
+              this.logger.verbose(`lock`, `updating lock file via "npm install --package-lock-only"`);
+              await exec('npm', ['install', '--package-lock-only'], { cwd: this.project.manifest.location });
+            } else {
+              // with npm, we need to do update the lock file in 2 steps
+              // 1. use shrinkwrap will delete current lock file and create new "npm-shrinkwrap.json" but will avoid npm retrieving package version info from registry
+              lockFilename = 'package-lock.json';
+              this.logger.verbose(`lock`, `updating lock file via "npm shrinkwrap --package-lock-only".`);
+              this.logger.warn(`npm`, `Your npm version is lower than 8.5.0, we strongly suggest to upgrade your npm client to avoid the use of "npm shrinkwrap" instead of the regular "npm install --package-lock-only".`);
+              await exec('npm', ['shrinkwrap', '--package-lock-only'], { cwd: this.project.manifest.location });
```
StarpTech, line 677: I'm not sure if `npm-shrinkwrap.json` is really backwards compatible with the `package-lock.json`. 8.5 is shipped with Node.js LTS. I'd show an error that users need to update to NPM >= 8.5.0 this is practical.

StarpTech: Oh yes they are identical https://docs.npmjs.com/cli/v8/configuring-npm/npm-shrinkwrap-json

StarpTech:
> I wonder if I should make this the new default, or just leave it as an opt-in flag since it might not work for everyone!?

I'd enable it by default but with an option to disable it, just in case. We are trying to fix an issue. I see no reason to enable it after some manual testing and unit tests.

lerna.json
ghiscoding: hmm but that is partially related to what I said in my last comment, I'm not sure if I'd make this new flag enabled by default. Also there's currently no logic in place to verify what type of workspace is in place, that's basically why there's an `npmClient` property in place and I'm not sure that I want to spend more time to deal with extra logic just to avoid the user to simply enable the flag himself. That topic actually came up in the other lock file opened PR, the solution by the other user was to use something like [preferred-pm](https://www.npmjs.com/package/preferred-pm)) to find out which is used but that package has a couple of dependencies and I'm not crazy about adding more and more dependencies... so is it worth the extra time

packages/core/src/models/command-options.ts
ghiscoding: `updateRootLockFile` is the flag to use when you want Lerna-Lite to update the lock file directly (what is currently in place for updating the npm lock file), while `packageLockfileOnly` would be the new flag in this PR. I'm considering keeping both approach but I agree that the flag `updateRootLockFile` is confusing and it could be renamed since it's not yet exposed in the version readme.

packages/version/src/version-command.ts
ghiscoding: I am on Node 16 and like I mentioned earlier, I was on 8.4.1 even though I'm on Node LTS, so I don't think this problem will disappear anytime soon and so the `logger.warn` does exactly that to display a warning and recommend to upgrade npm.

ghiscoding:
> > I wonder if I should make this the new default, or just leave it as an opt-in flag since it might not work for everyone!?
> 
> I'd enable it by default but with an option to disable it, just in case. We are trying to fix an already broken state. I see no reason to enable it after some manual testing and unit tests.

Not entirely broken, it is broken for pnpm yes but it's not currently broken for npm users and that is what that other flag is for `updateRootLockFile`, on a project that uses npm workspaces (like Lerna-Lite is), then current code has `updateRootLockFile` enabled by default and will update the lock file directly (in other words Lerna-Lite updates the file directly) and that is what the other opened PR #168 was trying to do. There are users that did tell that they use npm workspaces and rely on the other approach, they would probably not notice if I change which approach is use to update the lock file but I'd have to make sure it's always working (I wouldn't want to introduce a regression)

StarpTech:
> Since Lerna-Lite requires the user to have a workspace pre-setup, it might be ok to use this new flag and perhaps abandon the code to update the file directly which is always riskier and harder to deal with (you never know when any of the pm client change their lock file structure)

Make sense.

packages/core/src/models/command-options.ts
StarpTech: Got it, thanks. Renaming would be good.

packages/version/src/version-command.ts
StarpTech: I mean latest LTS. npm 8.5.0 was shipped with Node.js 16.15.0, Current LTS is 16.15.1

ghiscoding: sure but Node 16 LTS is only EOL on June 2024 and maybe another 6 months buffer then it's better to keep this `shrinkwrap` in place as a plan B for the foreseeable future haha. I also don't have the tendency to update very often Node myself so the plan B is a good thing. I'll consider this one resolved

ghiscoding:
@StarpTech 
I think I'm done with this PR, you can do a final review before I merge it (probably on Monday or Tuesday)

Note that I decided to **not** make this new flag the default because I'm afraid of possible breaking change (yarn 1.x classic will actually not work with it). I'm also afraid that some developers might be surprised that running a `version` command calls a client install lock only file and I also don't have a Yarn Berry setup (I assume it would but I don't have time to spend on installing/testing it)... so for all these reasons it's an opt-in flag and will ask developers to enable the new flag themselves. I did however add the `no-` flag for both options, so it can be enabled/disabled. Also Note that I found that the `--no-` flag can only work in the shell (meaning that `--no-package-lockfile-only` in the shell would negate the flag but adding `noPackageLockfileOnly` in `lerna.json` does not, however `packageLockfileOnly: false` should be ok).

I still have to test with an actual pnpm project, if you get a change to try yourself then that'd be great too. 

@JacobSoderblom if you have a chance to test this new approach, it would be nice too. You will need to enable the new flag in the shell or in `lerna.json` via `--package-lockfile-only` (this will run behind the scene `pnpm install --lockfile-only` and also add the updated lock file to Git changes)

StarpTech:
Awesome, thank you. I'll do and report. One thing. I find the name `packageLockfileOnly` confusing. It's clear from pnpm perspective but inside lerna.json it's not clear. I'd prefer something like `syncWorkspaceLock` or similar.

packages/version/src/version-command.ts
```diff
@@ -603,26 +615,78 @@ export class VersionCommand extends Command<VersionCommandOption> {
...
-      // update modern lockfile (version 2 or higher) when exist in the project root
-      loadPackageLockFileWhenExists(rootPath)
-        .then(lockFileResponse => {
-          if (lockFileResponse && lockFileResponse.lockfileVersion >= 2) {
-            for (const pkg of this.packagesToVersion) {
-              this.logger.silly(`lock`, `updating root "package-lock-json" for package "${pkg.name}"`);
-              updateTempModernLockfileVersion(pkg, lockFileResponse.json);
+    // update the project root lock file, we will read and write back to the lock file
+    // this is currently the default update and if none of the flag are enabled (or all undefined) then we'll consider this as enabled
+    if (this.options.manuallyUpdateRootLockfile || (this.options.manuallyUpdateRootLockfile === undefined && !this.options.packageLockfileOnly)) {
```
StarpTech, line 620: This is only executed for npm, right? Where is the check?

packages/version/src/version-command.ts
```diff
@@ -603,26 +615,78 @@ export class VersionCommand extends Command<VersionCommandOption> {
...
+            changedFiles.add(lockFilename);
+            break;
+          case 'npm':
+          default:
+            lockFilename = 'package-lock.json';
+            const localNpmVersion = execSync('npm', ['--version']);
+            this.logger.silly(`npm`, `current local npm version is "${localNpmVersion}"`);
+
+            // for npm version >=8.5.0 we can call "npm install --package-lock-only"
+            // when lower than we call "npm install --package-lock-only" and rename "npm-shrinkwrap.json" back to "package-lock.json"
```
StarpTech, line 669: I think you mean 

```
// when lower than we call "npm shrinkwrap --package-lock-only" and rename "npm-shrinkwrap.json" to "package-lock.json"
```

packages/version/src/__tests__/version-conventional-commits.spec.ts
```diff
@@ -228,7 +228,7 @@ describe("--conventional-commits", () => {
 
       const cwd = await initFixture("normal");
 
-      await new VersionCommand(createArgv(cwd, "--conventional-commits", "--conventional-prerelease"));
+      await new VersionCommand(createArgv(cwd, "--conventional-commits", "--conventional-prerelease", "--no-package-lockfile-only"));
```
StarpTech, line 231: Why is it necessary to add those flags to the tests?

StarpTech (changes requested):
Good state, but proper tests are missing.

packages/version/src/__tests__/version-conventional-commits.spec.ts
ghiscoding: because making my final decision I had switched it to the new flag and had to adjust a few unit tests in the process. I could remove them back now though or just leave them there

packages/version/src/version-command.ts
ghiscoding: fixed

packages/version/src/version-command.ts
ghiscoding: fixed

packages/version/src/__tests__/version-conventional-commits.spec.ts
StarpTech: Rmove them because it adds bloat to the unit test.

ghiscoding: fixed

ghiscoding:
Alright, I have addressed all comments identified in the review and I believe to be feature complete
- renamed `packageLockfileOnly` to `syncWorkspaceLock` (good suggestion)
- also decided to move the `install lockfile-only` logic execution out of the `VersionCommand` and move into `update-lockfile-version.ts` where it should be belong
- tested locally with Lerna-Lite with the npm option (lower and greater than npm version 8.5.0)

The only thing that I wanted to do was to test it with that `ts-pnpm-workspace` project but I don't understand how to use the `pnpm link` command. If I'd be using this with `npm link`, I would simply run the command from the `cli/dist` folder and I'd be good but that doesn't seem to work with `pnpm link`. How would you test with pnpm?

StarpTech:
> I used npm link in the past and I would simply run the command from the cli folder and I'd be good to go but that doesn't seem to work with pnpm link. How would you test with pnpm?

I don't use `pnpm link` directly. I use https://pnpm.io/pnpm-workspace_yaml

packages/version/src/version-command.ts
```diff
@@ -603,26 +614,46 @@ export class VersionCommand extends Command<VersionCommandOption> {
...
-              .then((lockfilePath) => {
-                if (lockfilePath) {
-                  changedFiles.add(lockfilePath);
-                }
-              });
-          }
-        })
-    );
+    // update lock file, with npm client defined when `--package-lock-only` is enabled
+    if (this.options.syncWorkspaceLock) {
```
StarpTech, line 647: I'd put this in the else branch of https://github.com/ghiscoding/lerna-lite/pull/199/files#diff-dce25c7083d885ad853224dab7020a62243a4c81d90ad681a3fad3f50b2adaa4R620

ghiscoding: no I specifically target the ones who use npm and falls under that other flag, I don't want to annoy other users with this message

ghiscoding:
> > I used npm link in the past and I would simply run the command from the cli folder and I'd be good to go but that doesn't seem to work with pnpm link. How would you test with pnpm?
> 
> I don't use `pnpm link` directly. I use https://pnpm.io/pnpm-workspace_yaml

I'm not exactly sure what you mean by this line, I mean if you had Lerna-Lite repo cloned, how would you link your pnpm project to use the local lerna-lite code? Typically an `npm link` would do the trick

packages/version/src/version-command.ts
StarpTech: I mean

```ts
if (npmClient === 'npm' && this.options.manuallyUpdateRootLockfile) {
 // ...
} else if (this.options.syncWorkspaceLock) {
 // ...
}
```

StarpTech:
> I'm not exactly sure what you mean by this line, I mean if you had Lerna-Lite repo cloned, how would you link your pnpm project to use the local lerna-lite code? Typically an npm link would do the trick

I misunderstood it. Yes, [`pnpm link`](https://pnpm.io/cli/link) should do the trick

ghiscoding:
@StarpTech 
This just got pushed under the new release [v1.5.0](https://github.com/ghiscoding/lerna-lite/releases/tag/v1.5.0) 🚀, could you also please let me know once you give it a try on your next version release. That would be nice to have a confirmation that it worked fine (at least I got confirmation from the npm side by releasing this new version, everything went smoothly but it would be nice to have confirmation with `workspace:` protocol in place, though I might migrate Lerna-Lite to pnpm in next few weeks). 

Thanks a lot for the great review and the great flag name suggestion 😉

StarpTech:
Hi @ghiscoding, will do and thanks for the hard work.

johnsoncodehk:
@ghiscoding I needed it for a lone time, thanks to implement this!

But I can't make it works with `lerna version` command, is `--no-git-tag-version` only for `lerna release`?

If you need reproduction, you can clone https://github.com/johnsoncodehk/volar and run `$ pnpm i && npm run version:test` to try.

ghiscoding:
@johnsoncodehk 
Hey thanks for using it, I did see you switched to Lerna-Lite not long ago and I'm also a user of Volar so it'd be great to help you. Just as a reminder, I did create Lerna-Lite fork but I don't know all of Lerna's original code (neither do I understand it all). It's probably better to track this under a new issue. I'd have to check if `--no-git-tag-version` has any impact (I didn't test without git tag so it's possible but I'm unsure). Would you mind opening a new issue for this though. Thanks

johnsoncodehk:
@ghiscoding Of course it's here: #214

(Btw `--no-git-tag-version` is typo, I copy wrong span, I mean `--sync-workspace-lock`)

ghiscoding:
quick investigation seems to point to a simple missing `npmClient`, we should be good. It still makes me want to migrate Lerna-Lite itself to pnpm even more. 😄

PR #272, opened 2022-07-24
feat(version): option to add commit login username on each changelog entry, closes #248

ghiscoding:
## Description

Add new `lerna version --changelog-include-commits-client-login [msg]` option that will add commit remote client login username to each changelog commit line entry. This option will fetch all commits from the client remote server (only GitHub for now) and from that list it will find associated user login with commit changelog entry and will add them at the end of each changelog entry (when found), it uses the [GitHub Graphql API](https://docs.github.com/en/graphql) to fetch all recent commits (since last release) from GitHub with only the fields we're interested in.

## Motivation and Context

closes #248, which is to add user login on each changelog commit line and that will indirectly create a Contributors list when creating a new GitHub Release

Example with default format
```md
**deps:** update dependency git-url-parse to v12 ([978bf36](https://github.com/.../978bf36)) (@renovate-bot)
```

or a custom format (use any of `%a` and/or `%l`)
```sh
lerna version --changelog-include-commits-client-login " by @%l, %a"

# result
**deps:** update dependency git-url-parse to v12 ([978bf36](https://github.com/.../978bf36)) by @renovate-bot, Whitesource Renovate
```
## How Has This Been Tested?

added necessary unit tests and tested in `git-dry-run` mode to see changelog being updated as intended

## Types of changes

- [ ] Chore (change that has absolutely no effect on users)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #315, opened 2022-08-24
fix(core): fix parsing commit date with different time zone

ahrbil:
## Description
Update regex for extracting commit date inside the `get-commits-since-last-release.ts` function, the regex has an issue related to parsing time zones. When executing `git log` with `--format="%h %aI"` option to get the oldest commit details,  [%aI](https://git-scm.com/docs/pretty-formats#Documentation/pretty-formats.txt-emaIem) formats the author date in ISO 8601 format, then the old regex `/^"?([0-9a-f]+)\s([0-9\-T\:]*)"?$/` used to extract commit hash and date. The old regex can only parse dates with a time zone offset of **- hours** but in my case, my time zone has an offset of **+ hours**. Hence, the commit date was always `undefined` leading to all commands that call Github API with $since variable to fail with GraphqlResponseError `Variable $since of type GitTimestamp! was provided invalid value`.

Thank you @ghiscoding for the fantastic work 🙏.

## Motivation and Context

After #272, I tried to use the --changelog-include-commits-client-login option but it always fails when calling GitHub API, after debugging turns out that the commit date was `undefined` because of miss-parsing of the ISO date time zones.

New Regex ([regex101 editor](https://regex101.com/r/8bRJUN/1))
```js
/^"?([0-9a-f]+)\s([0-9\-|\+T\:]*)"?$/
```
should parse -+ time zones.

## How Has This Been Tested?
All old tests pass and I don't know if this case needs a dedicated test.

## Types of changes

## Checklist:

### DIFF
--- a/packages/core/src/conventional-commits/__tests__/get-commits-since-last-release.spec.ts
+++ b/packages/core/src/conventional-commits/__tests__/get-commits-since-last-release.spec.ts
@@ -117,6 +117,10 @@ describe('getOldestCommitSinceLastTag', () => {
       });
     });
 
+    afterEach(() => {
+      (execSync as jest.Mock).mockReset();
+    });
+
     it('should expect a tag date & hash but queried with a particular tag match pattern when using independent mode', async () => {
       const isIndependent = true;
       const mockExecSyncResult = '"deadabcd 2022-07-01T00:01:02-06:00"';
@@ -135,10 +139,9 @@ describe('getOldestCommitSinceLastTag', () => {
 
     it('should expect a commit date and hash when using different time zone', async () => {
       const isIndependent = true;
+      (execSync as jest.Mock).mockReturnValue('"deadbeef 2022-07-01T00:01:02+01:00"');
       const result = await getOldestCommitSinceLastTag(execOpts, isIndependent, false);
-      const execSpy = (execSync as jest.Mock)
-        .mockReturnValueOnce('')
-        .mockReturnValueOnce('"deadbeef 2022-07-01T00:01:02+01:00"');
+      const execSpy = (execSync as jest.Mock).mockReturnValueOnce('"deadbeef 2022-07-01T00:01:02+01:00"');
 
       expect(describeRefSync).toHaveBeenCalledWith({ cwd: '/test', match: '*@*' }, false);
       expect(execSpy).toHaveBeenCalledWith(
--- a/packages/core/src/conventional-commits/__tests__/get-commits-since-last-release.spec.ts
+++ b/packages/core/src/conventional-commits/__tests__/get-commits-since-last-release.spec.ts
@@ -27,10 +27,10 @@ const commitsStub = [
 describe('getCommitsSinceLastRelease', () => {
   beforeEach(() => {
     (describeRefSync as jest.Mock).mockReturnValue(tagStub);
-    (execSync as jest.Mock).mockReturnValue('"deadbeef 2022-07-01T00:01:02-04:00"');
   });
 
   it('throws an error if used with a remote client other than "github"', async () => {
+    (execSync as jest.Mock).mockReturnValue('"deadbeef 2022-07-01T00:01:02-04:00"');
     await expect(getCommitsSinceLastRelease('gitlab', 'durable', 'main', false, execOpts)).rejects.toThrow(
       'Invalid remote client type, "github" is currently the only supported client with the option --changelog-include-commits-client-login.'
     );
@@ -38,6 +38,7 @@ describe('getCommitsSinceLastRelease', () => {
 
   it('should expect commits returned when using "github" when a valid tag is returned', async () => {
     (getGithubCommits as jest.Mock).mockResolvedValue(commitsStub);
+    (execSync as jest.Mock).mockReturnValue('"deadbeef 2022-07-01T00:01:02-04:00"');
     const isIndependent = false;
     const result = await getCommitsSinceLastRelease('github', 'durable', 'main', isIndependent, execOpts);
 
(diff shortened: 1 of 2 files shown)


ghiscoding:
oh thanks, I forgot where I took that regex from but that's totally a missed on my part. On the test side, could you just modify at least 1 of the current [unit tests](https://github.com/ghiscoding/lerna-lite/blob/main/packages/core/src/conventional-commits/__tests__/get-commits-since-last-release.spec.ts) and change one of them to be a `+` instead of a `-` and it should be enough to test the new regex

packages/core/src/conventional-commits/__tests__/get-commits-since-last-release.spec.ts
```diff
@@ -129,5 +129,21 @@ describe('getOldestCommitSinceLastTag', () => {
       );
       expect(result).toEqual({ commitDate: '2022-07-01T00:01:02-04:00', commitHash: 'deadbeef' });
     });
+
+    it('should expect a commit date and hash when using different time zone', async () => {
+      const isIndependent = true;
+      const result = await getOldestCommitSinceLastTag(execOpts, isIndependent, false);
+      const execSpy = (execSync as jest.Mock)
+        .mockReturnValueOnce('')
+        .mockReturnValueOnce('"deadbeef 2022-07-01T00:01:02+01:00"');
```
ahrbil, line 138: @ghiscoding is there a way to force the mocked returned value to "deadbeef 2022-07-01T00:01:02+01:00"? I still get the value "deadbeef  2022-07-01T00:01:02-04:00" passed.

ghiscoding: ah yeah I noticed that when I was writing these tests, I inadvertently used the same mocked result across all unit tests which isn`t correct, I just pushed an updated version of the tests just now, can you cherry pick my last commit into your branch and then modify it to your need. Thanks again

```sj
git cherry pick f618ae9
```

packages/core/src/conventional-commits/__tests__/get-commits-since-last-release.spec.ts
```diff
@@ -117,6 +117,10 @@ describe('getOldestCommitSinceLastTag', () => {
       });
     });
 
+    afterEach(() => {
```
ahrbil, line 120: @ghiscoding I got tests passed by mockReset `execSync` after each test.

packages/core/src/conventional-commits/__tests__/get-commits-since-last-release.spec.ts
ahrbil: That worked, Thank you a lot.

packages/core/src/conventional-commits/__tests__/get-commits-since-last-release.spec.ts
ghiscoding: great thanks, it still shows a conflict though, could you fix it? I modified the same unit tests as you so there's a conflict to resolve, thanks

ahrbil: Yes sure, it is now resolved 🙏 .

ghiscoding:
@ahrbil 
Thanks again for the PR, I'll release a patch version later tonight after working hours. Cheers

ahrbil:
Currently migrating away from Lerna to Lerna-lite with pnpm and turborepo, but I got blocked by this issue. Thank you @ghiscoding for your valuable time and work.

ghiscoding:
Very welcome, I'm also concerned about some changes they've done in Lerna by the Nrwl team. It's really nice to see that Lerna is back but at the same time Nrwl is pushing so much for their Nx integration (they recently added Nx as a [dependency](https://github.com/lerna/lerna/blob/main/core/lerna/package.json#L65) of Lerna, but why??) which is not always great for users who don't want Nx (like users of TurboRepo like you). Some in some ways, it pushes me to keep Lerna-Lite going 😉

ghiscoding:
published patch release [v1.11.1](https://github.com/ghiscoding/lerna-lite/releases/tag/v1.11.1) (lots of 1s there, must be a lucky release 😄)

ahrbil:
I agree @ghiscoding, I think Lerna now will be more associated with Nx. The reason I'm migrating from Lerna is pnpm support and `workspace:` protocol, Nx team is working on that at the moment, and this will bring us to the second reason which is performance, Lerna used to be an all-in-one tool that does everything but nowadays we have dedicated tools that only specialize at one thing and does it very well, turborepo is much performant than Lerna `exec` and `run`, pnpm linking and resolving of workspace packages replaces Lerna `add` and `bootstrap`, sure there will be more configuration but this is something you don't deal with every day. The missing part for my migration was versioning and generating  changelog from conventional commits, sure there is [changesets](https://github.com/changesets/changesets) but I think it is more dedicated to libraries, and also I like this command from Lerna and it's the most command I use, I was looking for alternatives but none of them were straightforward. I already heard about `lerna-lite` but back then doesn't have pnpm support, now the puzzle is solved with pnpm support landing.

ahrbil:
> published patch release [v1.11.1](https://github.com/ghiscoding/lerna-lite/releases/tag/v1.11.1) (lots of 1s there, must be a lucky release 😄)

You need to hear the fun part, it is also my first contribution to an open source project 😆.

ghiscoding:
Awesome, welcome to the Open Source world 😉 and feel free to reach out if it still doesn't work. I also agree with everything you said.

>  I already heard about lerna-lite but back then doesn't have pnpm support, now the puzzle is solved with pnpm support landing.

I added support for it about 3-4 months ago and I converted the project to pnpm workspace protocol about a month later. After a couple of small fixes around the feature, it's quite stable now, The biggest player now using this feature with Lerna-Lite is [Jest](https://github.com/facebook/jest), which I was thrilled to see them migrate just a couple weeks ago 😸

PR #316, opened 2022-08-24
WIP - chore: use pnpm `shell-emulator` instead of pre/post to cleanup scripts

ghiscoding:
## Description

cleanup pnpm scripts

## Motivation and Context

the `shell-emulator` in `.npmrc` allows us to have cross-platform npm scripts and as pnpm pointed out in their docs, it better to be implicit and avoid pre/post scripts because they are less visible to the users and it avoids surprises

## Types of changes

## Checklist:


ghiscoding:
running prettier and tsc fails, it seems to have problems with the paths, opened an issue on the pnpm project [issue 5260](https://github.com/pnpm/pnpm/issues/5260), at this point it doesn't look like I could use this. Another alternative would be to use `npm-run-all` but I would rather not add extra dependencies

PR #410, opened 2022-11-09
feat: make npm-publish to override publishConfig

artechventure:
## Description

Make npm-publish function to ovveride manifest by publishConfig.

## Motivation and Context

https://github.com/ghiscoding/lerna-lite/issues/404

As described in an issue above, yarn and pnpm has it's own logic to overwrite publishConfig to manifest.
So I added that logic into npm-publish function.
Added ones doesn't effect existing publishConfig related logics.

## How Has This Been Tested?

Added an test code.

## Types of changes

## Checklist:

### DIFF
--- a/packages/publish/src/lib/pack-directory.ts
+++ b/packages/publish/src/lib/pack-directory.ts
@@ -3,6 +3,7 @@ import path from 'path';
 import packlist from 'npm-packlist';
 import log from 'npmlog';
 import tar from 'tar';
+import fs from 'fs-extra';
 
 import { LifecycleConfig, Package, PackConfig, runLifecycle, tempWrite } from '@lerna-lite/core';
 import { getPacked } from './get-packed';
@@ -41,6 +42,9 @@ export async function packDirectory(_pkg: Package, dir: string, options: PackCon
   await runLifecycle(pkg, 'prepack', opts);
   await pkg.refresh();
 
+  await overwritePublishConfig(pkg.manifestLocation);
+  await pkg.refresh();
+
   const arborist = new Arborist({ path: pkg.contents });
   const tree = await arborist.loadActual();
   const files: string[] = await packlist(tree);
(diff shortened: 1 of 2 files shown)


packages/publish/src/lib/pack-directory.ts
```diff
@@ -69,6 +73,47 @@ export async function packDirectory(_pkg: Package, dir: string, options: PackCon
...
+  'browser',
+  'esnext',
+  'es2015',
+  'unpkg',
+  'umd:main',
+  'cpu',
+  'os',
+];
+
+async function overwritePublishConfig(manifestLocation: string) {
```
ghiscoding, line 92: I would prefer "override" instead of "overwrite" because the latter can be confused by thinking it's overwrites the entire file which is far from it. Also, pnpm wrote the something similar in their docs

> It is possible to override some fields in the manifest before the package is packed. The following fields may be overridden:

packages/publish/src/lib/pack-directory.ts
```diff
@@ -69,6 +73,47 @@ export async function packDirectory(_pkg: Package, dir: string, options: PackCon
   );
 }
 
+export const PUBLISH_CONFIG_FIELDS = [
```
ghiscoding, line 76: it looks like there's also a "typesVersion" that pnpm can also override as defined in their docs
https://pnpm.io/package_json#publishconfig

packages/publish/src/lib/pack-directory.ts
```diff
@@ -69,6 +73,47 @@ export async function packDirectory(_pkg: Package, dir: string, options: PackCon
...
+  if (manifest.publishConfig === undefined) {
+    return;
+  }
+
+  const publishConfig = { ...manifest.publishConfig };
+
+  PUBLISH_CONFIG_FIELDS.forEach((key) => {
+    if (key in publishConfig) {
+      manifest[key] = publishConfig[key];
+      delete publishConfig[key];
```
ghiscoding, line 104: I didn't test the code but I wonder what's the reason that you have to delete the object property? I see you're also deleting a few lines below as well

ghiscoding (changes requested):

artechventure:
Thanks for your comment. I think providing through options flag is a good idea. I will reflect you reviews this weekend. Thanks!

ghiscoding:
@artechventure I had a chance to take a look at your PR today. By looking at the PR and trying to add unit tests, I discovered that actually the code you added should really be moved into the `publish-command.ts` and `package.ts` instead of `pack-directory`, the main reason is because we do all temp publish changes into `publish-command.ts` just before publishing, this way it keeps the source intact and the temp changes only exist when publishing (we do the same for replacing `workspace:` protocol, and `links:` deps, so this override becomes just another step in that list of things to do before publishing as you can see below)

https://github.com/ghiscoding/lerna-lite/blob/67705c54e942d0873d081e755992e6bdeebf344a/packages/publish/src/publish-command.ts#L261-L282

I'll be closing this PR, but don't worry you will still be a contributor since I forked your PR 😉 
Could you please review the new PR #415 that I just created to supersede yours. 
Thanks for the contributions

PR #484, opened 2023-02-17
chore: bump jest version

mrazauskas:
## Description

https://github.com/facebook/jest/issues/13925#issuecomment-1433522739

## Motivation and Context

An issue with `jest.restoreAllMocks()` usage prevents upgrading Jest. This PR solve the issue.

## How Has This Been Tested?

The failing test is now passing locally.

## Types of changes

## Checklist:

## Summary

In general my first idea was correct.

There is no need to have `jest.clearAllMocks()` in the test. It has no effect in this case, because mocks are cleared (e.g. the count of calls is set to 0) before each test through a setting in config:

https://github.com/lerna-lite/lerna-lite/blob/aa35a81736930755d71b3948cdf777da14defa86/jest/jest.config.ts#L5

The problem with `jest.restoreAllMocks()` is that it removes implementation of all mocks. It is called in `afterEach`, so there is no effect for the first test. The following tests fail, because this mock has no implementation anymore:

https://github.com/lerna-lite/lerna-lite/blob/aa35a81736930755d71b3948cdf777da14defa86/packages/version/src/__tests__/git-push.spec.ts#L16

The `mockImplementationOnce` works as expected. The thing is that `exec` is called twice in tests with `mockImplementationOnce`. So first call has single use implementation, but second call is performed on reset mock with no implementation. Tests are expecting two calls:

https://github.com/lerna-lite/lerna-lite/blob/aa35a81736930755d71b3948cdf777da14defa86/packages/version/src/__tests__/git-push.spec.ts#L71

Also looking at [source](https://github.com/lerna-lite/lerna-lite/blob/main/packages/version/src/lib/git-push.ts) of `gitPush()` (the function under test), one can see that `exec` will be called twice if an error occurs.


mrazauskas:
Hm.. Not sure if I did everything right with `pnpm`. Not using it. So this part is just a guess.

ghiscoding:
@mrazauskas wow thanks a lot for the PR, really appreciate your time spent on this. Have a nice day :)

Issue #510, opened 2023-03-10
Sign multiple tags fails


heilmela:
### Describe the bug

I have a monorepo with (~ 16) packages and individual  versionin. Running `lerna version --exact --conventional-commits --sign-git-commit --sign-git-tag --tag-version-prefix=''` leads to:
```shell
lerna-lite ERR! lerna-lite Command failed with exit code 128: git tag packageX@1.6.0 -m packageX@1.6.0 --sign
lerna-lite ERR! lerna-lite error: gpg failed to sign the data
lerna-lite ERR! lerna-lite error: unable to sign the tag
```
This Errors does not occur when only a few package are bumped and i can individually execute git tag ... --sign myself. So it doesnt seem to be a gpg config issue.

### Expectation

Signing of many tags should work

### Reproduction

Create a monorepo with 16 packages and bump the version of all with 
```shell
lerna version --exact --conventional-commits --sign-git-commit --sign-git-tag --tag-version-prefix=''
```
The error does not occur every time, it seems to be some kind of timing issue

### Lerna config and logs

<details><summary>lerna.json</summary><p>

```json
{
  "packages": ["packages/*", "apps/*"],
  "version": "independent",
  "npmClient": "yarn",
  "useWorkspaces": true,
  "command": {
    "version": {
      "allowBranch": "main",
      "message": "chore: published new release"
    }
  }
}

```
</p></details>

### Environment Info

yarn 3.2.1
node v16.13.0
@lerna-lite/cli  1.16.0
MacOS (intel)

### Used Package Manager

yarn >= 2.x

### Validations

- [X] Follow our [Code of Conduct](https://github.com/lerna-lite/lerna-lite/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://github.com/lerna-lite/lerna-lite#readme).
- [X] Check that there isn't [already an issue](https://github.com/lerna-lite/lerna-lite/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/lerna-lite/lerna-lite/discussions).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #511, opened 2023-03-10
fix(version): sign tags sequentially to prevent gpg issues with many tags

heilmela:
## Description

Tags are signed sequentially 

## Motivation and Context

fixes https://github.com/lerna-lite/lerna-lite/issues/510#issue-1619420415

Simultaneously executing `git tag ... --sign` seems to be an issue. I still have no clue why exactly gpg starts to fail.

On the other hand sequentially signing even with 16 tags has a barely noticeable difference in speed compared to simultanously executing a sign requests.

## How Has This Been Tested?

Unfortunately the issue seems to be a timing issue with gpg, thus i repeated signing 10 times of which non failed.

## Types of changes

## Checklist:


ghiscoding:
Thanks for the fix, that code was implemented by another user but your fix seems good even if it now changes it be executed in series. Thanks for the contribution

ghiscoding:
release under [v1.16.1](https://github.com/lerna-lite/lerna-lite/releases)

PR #537, opened 2023-04-04
feat(build)!: migrate to ESM build & switch to Vitest for ESM support

ghiscoding:
## Description

Breaking Change: Migrate from CommonJS to ESM TypeScript build and migrate from Jest to Vitest since it has better ESM support

## Motivation and Context

- switch to ESM build for much smaller build output (instead of CommonJS which is currently used)
- upgrade all npm dependencies to latest including deps that are ESM only deps (also remove skipped exclusions from Renovate)
- migrate from Jest to Vitest since the later supports ESM by default (Jest seems far away from supporting ESM)

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:


ghiscoding:
will merge with failures so that we can keep a ref to troubleshoot Vitest Snapshot failures in opened Vitest [issue](https://github.com/vitest-dev/vitest/issues/3129)

PR #573, opened 2023-04-15
fix: cannot find `publish` module if `cli` inside a monorepo package

johnsoncodehk:
## Description

Fixes `Cannot find package '@lerna-lite/publish' imported from ...` error for monorepo projects.

## How Has This Been Tested?

- Clone https://github.com/volarjs/workspace
- Run `$ pnpm i && npm run setup:vue`
- Run `$ cd volar && npm run release`

## Types of changes

## Checklist:

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -62,6 +62,14 @@
     "npmlog": "^7.0.1",
     "yargs": "^17.7.1"
   },
+  "peerDependencies": {
+    "@lerna-lite/publish": "2.0.0"
+  },
+  "peerDependenciesMeta": {
+    "@lerna-lite/publish": {
+      "optional": true
+    }
+  },
   "devDependencies": {
     "yargs-parser": "^21.1.1"
   }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -188,6 +188,9 @@ importers:
       '@lerna-lite/init':
         specifier: workspace:*
         version: link:../init
+      '@lerna-lite/publish':
+        specifier: 2.0.0
+        version: link:../publish
       dedent:
         specifier: ^0.7.0
         version: 0.7.0


ghiscoding:
@johnsoncodehk thanks for the contribution, it looks like the PR is failing because the pnpm lock file is outdated. I also wonder why you got this error and you didn't have this problem before with version 1.x. Are you having this problem because you mark @lerna-lite as an `optionalDependencies`? All I could find is this Stack Overflow, https://stackoverflow.com/a/66228639/1212166 and I guess it is indeed because of the optional deps, just want to confirm. Thanks

Note that I upgraded to pnpm v8 today, if you are still on v7 and need help to update the lock file then let know and I can update by forking your PR, they changed the lock file structure quite a lot in v8 (see pnpm [v8](https://github.com/pnpm/pnpm/releases/tag/v8.0.0))

johnsoncodehk:
Your understanding of `peerDependenciesMeta` is correct, the use of peerDependency is for `publish` to be promoted to a location that `cli` can access when using monorepo + pnpm, but there seem still have other similar problems.

I'll try to provide a minimal reproduction tomorrow.

ghiscoding:
wow do you ever sleep? go to bed, there's no rush 😉

ghiscoding:
@johnsoncodehk is the PR still required since I couldn't replicate with pnpm 8 (as per my previous comment)?

johnsoncodehk:
Sorry for the delay, I will find time to check today.

johnsoncodehk:
This is a minimal reproduction: https://github.com/johnsoncodehk/lerna-lite-573

ghiscoding:
@johnsoncodehk thanks for the repro, I tried it and do see the error. What I wonder however is this, do we need this change in other package as well? I'm asking because in this case you are only interested in `publish` since `version` is a dependency anyway but what about other commands (or even `version` installed alone without `publish`), will they suffer from the same problem and should we do this modif for all package? I'm not too familiar with this `optionalDependencies` so your input would be helpful

johnsoncodehk:
Yes you should make the same changes to all related packages, if you agree I will close the PR and leave it to you.

ghiscoding:
@johnsoncodehk yeah that's fine, I will merge your PR and contribution and apply this change to all packages later. I assume you can wait few more days right? I mean this is not a blocker is it? Because I have some small CLI issues to resolve with another user and it might take couple days since we aren't on the same TZ (like you and me too ;)

johnsoncodehk:
Oh thanks. :)

ghiscoding:
ok thanks, I'll ping you on the next release. Thanks for the contribution :)

Could you review PR #578 that I just opened for all optional commands. I also changed the peer to `2.x` since it seems valid as per NodeJS [docs](https://nodejs.org/en/blog/npm/peer-dependencies#using-peer-dependencies) and I don't want to update the peer deps every time, unless you think it should be fixed to each Lerna-Lite version release?

johnsoncodehk:
`2.x` is good enough, even `*` fine IMO.

Issue #569, opened 2023-04-14
Versioning: `premajor` broken and doesn't match `--dry-run`


Jason3S:
### Describe the bug

Dry Run and Actual results in different versions:

Command:
```sh
lerna version --conventional-commits --sync-workspace-lock --yes --conventional-prerelease --dry-run premajor
```

Version: `6.31.1 => 7.0.0-alpha.0`

<img width="939" alt="image" src="https://user-images.githubusercontent.com/3740137/232016620-1843c63b-e611-4f40-9781-28a633ef671f.png">

Command:
```sh
lerna version --conventional-commits --sync-workspace-lock --yes --conventional-prerelease premajor
```

Version: `6.31.1 => 6.31.2`

<img width="882" alt="image" src="https://user-images.githubusercontent.com/3740137/232017487-7b76a763-86bc-4969-bfdc-ec6e51bd069d.png">

### Expectation

Expect that the version generated without `--dry-run` matches the dry-run version.

### Reproduction

There seems to be two different issues here:
1. `--dry-run` and non-dry-run do not match.
2. `premajor` and or `--conventional-prerelease` do not seem to be working.

If needed, I'll make a minimal reproduction, but given the obvious issue here, it is already something to be handled by the Unit Tests.

### Lerna config and logs

<details><summary>lerna.json</summary><p>

```json
{
  "npmClient": "pnpm",
  "packages": [
    "./packages/*",
    "./rfc/*",
    "./test-packages/*/*",
    "./integration-tests"
  ],
  "version": "6.31.2",
  "command": {
    "version": {
      "allowBranch": "main",
      "conventionalCommits": true
    }
  }
}
```
</p></details>

### Environment Info

```
System:
    OS: Linux 5.15 Ubuntu 22.04.2 LTS 22.04.2 LTS (Jammy Jellyfish)
    CPU: (2) x64 Intel(R) Xeon(R) CPU E5-2673 v3 @ 2.40GHz
    Memory: 5.87 GB / 6.78 GB
    Container: Yes
    Shell: 5.1.16 - /bin/bash
  Binaries:
    Node: 16.20.0 - /opt/hostedtoolcache/node/16.20.0/x64/bin/node
    Yarn: 1.22.[19](https://github.com/streetsidesoftware/cspell/actions/runs/4699014859/jobs/8331987269#step:13:20) - /usr/local/bin/yarn
[...]

PR #576, opened 2023-04-18
test: Add parsing test for version cli args

Jason3S:
## Description
Add Unit tests to verify the command line is parsed as expected.

## Motivation and Context

Needed to check exactly how `yargs` parsed the command line. Related to #569 and #575 

## How Has This Been Tested?

It is only a test.

## Types of changes

## Checklist:

### DIFF
--- a/packages/cli/src/cli-commands/__tests__/cli-version-commands.spec.ts
+++ b/packages/cli/src/cli-commands/__tests__/cli-version-commands.spec.ts
@@ -2,6 +2,9 @@ import { describe, expect, it, vi } from 'vitest';
 
 vi.mock('@lerna-lite/version', () => null);
 import cliVersion from '../cli-version-commands';
+import yargs from 'yargs/yargs';
+
+const oc = expect.objectContaining;
 
 describe('Version Command CLI options', () => {
   it('should log a console error when versionCommand is not provided', async () => {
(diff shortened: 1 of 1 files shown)


Jason3S:
@ghiscoding,

If there isn't some strange side-effect I was not expected, I don't believe the errors have anything to do with me adding some more unit test. Might be related to this being a PR from a fork.

Jason3S:
@ghiscoding,

This PR was only to demonstrate how I tested my assumptions about how the command line was parsed.

ghiscoding:
yeah I saw that too, I'm not sure why this is happening considering that `version` is mocked to return `null` and that is why I added my unit tests directly in the `version-command.spec.ts` in my other PR. It might be the move to Vitest that caused this, I don't know. I could take a look later after work. I just got 20min to check some other things over lunch but now it's over... I left a comment & question in the PR I've opened if you could review and reply. Thanks

https://github.com/lerna-lite/lerna-lite/blob/60afb34009b862ea4f3f4a02c80c0c298076c16a/packages/cli/src/cli-commands/__tests__/cli-version-commands.spec.ts#L3-L4

packages/cli/src/cli-commands/__tests__/cli-version-commands.spec.ts
```diff
@@ -2,6 +2,9 @@ import { describe, expect, it, vi } from 'vitest';
 
 vi.mock('@lerna-lite/version', () => null);
```
ghiscoding, line 3: ok I got all your tests working by changing the mock, so change the line 3
```ts
vi.mock('@lerna-lite/version', () => null);
```

with this and that should work
```ts
vi.mock('@lerna-lite/version', () => {
  const VersionCommand = null;
  return { VersionCommand };
});
```
image

ghiscoding:
So I ended up splitting the first PR #575 I had to keep only `conventional-prerelease` code change (without your suggestion) as a new PR #577. If you think that `--dry-run` should still be covered then feel free to add code for that in this PR or a separate PR (if you do, then you probably want to revisit PR #575 to copy over the unit tests I had for it)

PR #578, opened 2023-04-19
fix: cannot find optional commands if `cli` inside a monorepo package

ghiscoding:
## Description

Spreading the previous PR #573 to all command to fix the same problem but for all commands when installed as `optionalDependencies` inside a monorepo that contains a monorepo

## Motivation and Context

As per PR #573, this fixes `Cannot find package '@lerna-lite/publish' imported from ...` error for monorepo projects when any of the commands are installed as `optionalDependencies` (like [Volar](https://github.com/vuejs/language-tools) does)

## Types of changes

## Checklist:

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -62,15 +62,35 @@
     "npmlog": "^7.0.1",
     "yargs": "^17.7.1"
   },
-  "peerDependencies": {
-    "@lerna-lite/publish": "2.0.0"
-  },
-  "peerDependenciesMeta": {
-    "@lerna-lite/publish": {
-      "optional": true
-    }
-  },
   "devDependencies": {
     "yargs-parser": "^21.1.1"
+  },
+  "peerDependencies": {
+    "@lerna-lite/exec": "2.x",
+    "@lerna-lite/list": "2.x",
+    "@lerna-lite/publish": "2.x",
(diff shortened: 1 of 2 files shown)


ghiscoding:
@johnsoncodehk would you mind making a quick review of the PR, thanks

johnsoncodehk:
LGTM. 👍

ghiscoding:
@johnsoncodehk I'm having issues with this change, it breaks the publish process when I tried to push a new Lerna-Lite release. I had to test the rollback of every PRs to find that it is this PR causing the publish error. 

It seems that npm arborist has problem building the tree structure of the CLI package.json, which is basically this change. If I rollback the change I can then run a publish previous (dry-run) without any problems. I'm going to test couple more things but I might have to rollback this change. Do you know if there's anything that could cause this problem?

image

ghiscoding:
@johnsoncodehk 
I went ahead with only the `peerDependenciesMeta` (in PR #585) so that I was able to go ahead and publish a new release. Hopefully this works for you even without the `peerDependencies`, otherwise I'm not sure what else to do, there are similar issues in NPM (like https://github.com/npm/cli/issues/3847) but I think it will be ok since I see this issue/comment in pnpm [Make peerDependenciesMeta imply peerDependencies](https://github.com/pnpm/pnpm/issues/2128)

PR #579, opened 2023-04-19
fix(publish): add normalize-path as a dependency

rfoel:
## Description

`normalize-path` should be a dependency of the `@lerna-lite/publish` package, otherwise it throws errors while running on CI.

## Motivation and Context

While running a workflow on GitHub Actions, one of my packages wasn't published because the package `normalize-path` was missing. Installing it as a dependency of my repository worked, but as it is a dependency of `@lerna-lite/publish` it should be installed in the package itself instead.

image

## How Has This Been Tested?

No testing needed.

## Types of changes

## Checklist:

### DIFF
--- a/packages/publish/package.json
+++ b/packages/publish/package.json
@@ -61,6 +61,7 @@
     "has-unicode": "^2.0.1",
     "libnpmaccess": "^7.0.2",
     "libnpmpublish": "^7.1.3",
+    "normalize-path": "^3.0.0",
     "npm-package-arg": "^10.1.0",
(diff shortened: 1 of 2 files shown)


ghiscoding:
oh wow thanks for that, I'm surprised that this issue wasn't brought up earlier. Note that it might take a few more days to do a release, there's another PR to resolve before then. Thanks a lot for the fix

ghiscoding:
Now released under [v2.1.0](https://github.com/lerna-lite/lerna-lite/releases/tag/v2.1.0), thanks a lot for the contribution

PR #591, opened 2023-04-27
fix(version): allowPeerDependenciesUpdate should work w/`workspace:^`, fix #590

### LINKED ISSUE #590: Handling Workspace Protocol in Peer Dependencies with Lerna
### Discussed in https://github.com/lerna-lite/lerna-lite/discussions/587

<div type='discussions-op-text'>

<sup>Originally posted by **ma-multipla** April 25, 2023</sup>
I'm having trouble comprehending the purpose of a specific code fragment. In the context of utilizing the workspace protocol for peer dependencies (e.g. `"workspace:^"`). Yarn Berry's default behavior is to substitute them with the locally resolved version upon generating the tarball through `yarn pack`. However, with the current Lerna version, the workspace protocol remains unaltered when applied to peer dependencies, resulting in incompatibility with package managers that don't support the protocol, such as npm.

Using the `--allow-peer-dependencies-update` flag only bumps the version in specific cases and in my case it only removed the "workspace:" portion, leaving the "^".  I've read the documentation about the flag, but I have difficulties seeing the benefits of the current approach.
My questions is as follows:

Why isn't the `--allow-peer-dependencies-update` flag enabled by default for the workspace protocol and why is it only bumping the workspace protocol when the version is specified like this `"workspace:^1.2.0"`. This essentially makes it not possible to only use the npm range operator without specifying the version, e.g. `"workspace:^"`. Why would you want this? 
The documentation is also a bit misleading: 
```json
{
  "name": "B",
  "dependencies": {
    "A": "workspace:^1.2.0"   // will update to "workspace:^1.3.0",
    "B": "^0.4.0":            // will update to "^0.5.0"
   },
  "peerDependencies": {
    "A": "workspace:^1.2.0"   // will update to "workspace:^1.3.0"
    "B": ">=0.2.0":           // will not be updateed because range with operator (>=) are skipped
  }
}
```
https://github.com/lerna-lite/lerna-lite/blob/main/packages/version/README.md#with-flag-enabled

It will update to "^1.3.0" not "workspace:^1.3.0". My expectation would be that either `"workspace:^"` as well as `"workspace:^1.2.0"` should lead to a version bump to `"^1.3.0"` considering the example above.

Here's the code snippet in question:
```js
// Perform certain actions when there are peer dependencies
if (this.peerDependencies?.[depName]) {
  // If the user allows peer bump and it's a regular semver version, add it to the array of dependencies to potentially bump
  // However, don't do this when the semver has a range with an operator, e.g., bump ("^2.0.0") but not (">=2.0.0" or "workspace:<2.0.0" or "workspace:*")
  // prettier-ignore
  if (allowPeerDependenciesUpdate && /^(workspace:)?[~^]?[\d\.]+([\-]+[\w\.\-\+]+)*$/i.test(this.peerDependencies[depName] || '')) {
    updatingDependencies.push(this.peerDependencies);
  }
  // When peer bump is disabled, some peerDependencies might not be reviewed and could still have the `workspace:` prefix, so ensure to remove any of these prefixes
  else if (updatedByCommand === 'publish' && this.peerDependencies[depName].startsWith('workspace:')) {
    this.peerDependencies[depName] = this.peerDependencies[depName].replace('workspace:', '');
  }
}
```
Source: [Lerna Lite Core Package.ts](https://github.com/lerna-lite/lerna-lite/blob/2aed0c016f0805d44dc1607c20a5226c4b6886f1/packages/core/src/package.ts#L314)</div>

ghiscoding:
## Description

The `peerDependencies` is a special case and will only be bumped when the flag `--allow-peer-dependencies-update` is enabled. A package set as `workspace:^` (without any semver number) should be able to be bumped and it wasn't before this PR.

## Motivation and Context

The `--allow-peer-dependencies-update` implementation was only tested with `workspace:` and semver, but it wasn't tested with just `workspace:^` (or `~`) and the end result was actually producing an invalid version `^` (which is obviously wrong). 

This PR fixes 2 problems with `peerDependencies` identified in #587 and #590:

_the example below is assuming that the version is at `1.0.0` and we are processing a `minor` bump_
1. when version is `workspace:^` and `--allow-peer-dependencies-update` is enabled, the version in `package.json` should remain as `workspace:^` but the version that will be published should be bumped to `"1.1.0"`
   - this issue was caused by a very small RegEx that wasn't being match when using `workspace:^` without a semver number
3. when version is `workspace:^` and we **do not** provide `--allow-peer-dependencies-update`, it was previously trying to publish as `^` (which is an invalid version). In this case we now fallback to the previously resolved version when publishing (in this case `"1.0.0"`)

## Types of changes

## Checklist:

### DIFF
--- a/packages/core/src/package.ts
+++ b/packages/core/src/package.ts
@@ -311,7 +311,7 @@ export class Package {
       // when user allows peer bump and is a regular semver version, we'll push it to the array of dependencies to potentially bump
       // however we won't when the semver has a range with operator, ie this would bump ("^2.0.0") but the following would not (">=2.0.0" or "workspace:<2.0.0" or "workspace:*")
       // prettier-ignore
-      if (allowPeerDependenciesUpdate && /^(workspace:)?[~^]?[\d\.]*([\-]+[\w\.\-\+]+)*$/i.test(this.peerDependencies[depName] || '')) {
+      if (allowPeerDependenciesUpdate && /^(workspace:)?[~^*]?[\d\.]*([\-]+[\w\.\-\+]+)*$/i.test(this.peerDependencies[depName] || '')) {
         updatingDependencies.push(this.peerDependencies);
       }
       // when peer bump is disabled, we could end up with peerDependencies not being reviewed
--- a/packages/core/src/package.ts
+++ b/packages/core/src/package.ts
@@ -309,7 +309,7 @@ export class Package {
     // when we have peer dependencies, we might need to perform certain actions
     if (this.peerDependencies?.[depName]) {
       // when user allows peer bump and is a regular semver version, we'll push it to the array of dependencies to potentially bump
-      // however we won't when the semver has a range with operator, ie this would bump ("^2.0.0") but the following would not (">=2.0.0" or "workspace:<2.0.0" or "workspace:*")
+      // however we won't when the semver has a range with operator, ie this would bump ("^2.0.0") but the following would not (">=2.0.0", "14 || 15" or "workspace:<2.0.0")
       // prettier-ignore
       if (allowPeerDependenciesUpdate && /^(workspace:)?[~^*]?[\d\.]*([\-]+[\w\.\-\+]+)*$/i.test(this.peerDependencies[depName] || '')) {
         updatingDependencies.push(this.peerDependencies);
--- a/packages/core/src/package.ts
+++ b/packages/core/src/package.ts
@@ -306,18 +306,24 @@ export class Package {
     const localDependencies = this.retrievePackageDependencies(depName);
     const updatingDependencies = [localDependencies];
 
-    // when we have peer dependencies, we might need to perform certain things
+    // when we have peer dependencies, we might need to perform certain actions
     if (this.peerDependencies?.[depName]) {
       // when user allows peer bump and is a regular semver version, we'll push it to the array of dependencies to potentially bump
-      // however we won't when the semver has a range with operator, ie this would bump ("^2.0.0") but these would not (">=2.0.0" or "workspace:<2.0.0" or "workspace:*")
+      // however we won't when the semver has a range with operator, ie this would bump ("^2.0.0") but the following would not (">=2.0.0" or "workspace:<2.0.0" or "workspace:*")
       // prettier-ignore
-      if (allowPeerDependenciesUpdate && /^(workspace:)?[~^]?[\d\.]+([\-]+[\w\.\-\+]+)*$/i.test(this.peerDependencies[depName] || '')) {
+      if (allowPeerDependenciesUpdate && /^(workspace:)?[~^]?[\d\.]*([\-]+[\w\.\-\+]+)*$/i.test(this.peerDependencies[depName] || '')) {
         updatingDependencies.push(this.peerDependencies);
       }
       // when peer bump is disabled, we could end up with peerDependencies not being reviewed
       // and some might still have the `workspace:` prefix so make sure to remove any of these prefixes
       else if (updatedByCommand === 'publish' && this.peerDependencies[depName].startsWith('workspace:')) {
         this.peerDependencies[depName] = this.peerDependencies[depName].replace('workspace:', '');
+
+        // when it's only 1 char left "^" or "~", we'll assume that the version is invalid (note that "*" is valid)
+        // so reassigning the resolved package version seems like the best action to do in this case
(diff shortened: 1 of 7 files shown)


packages/core/src/package.ts
```diff
@@ -306,18 +306,24 @@ export class Package {
     const localDependencies = this.retrievePackageDependencies(depName);
     const updatingDependencies = [localDependencies];
 
-    // when we have peer dependencies, we might need to perform certain things
+    // when we have peer dependencies, we might need to perform certain actions
     if (this.peerDependencies?.[depName]) {
       // when user allows peer bump and is a regular semver version, we'll push it to the array of dependencies to potentially bump
-      // however we won't when the semver has a range with operator, ie this would bump ("^2.0.0") but these would not (">=2.0.0" or "workspace:<2.0.0" or "workspace:*")
+      // however we won't when the semver has a range with operator, ie this would bump ("^2.0.0") but the following would not (">=2.0.0" or "workspace:<2.0.0" or "workspace:*")
```
ghiscoding, line 312: actually re-reading the comments I left and if we now support `workspace:^` (and `~`) then we might as well support `workspace:*` too (which would equal to current version)

ghiscoding: done

ma-multipla:
This looks great! What might be confusing is the inconsistency to `yarn pack` - not sure how pnpm is currently handling this. When using `workspace:*` it would be bumped to the locally resolved version as exact version (no range operator) and with this change it will resolve `"*"`, this is a huge difference to the native handling. Might be sufficient to document this.

packages/publish/src/__tests__/publish-workspace-protocol-specifiers.spec.ts
```diff
@@ -97,9 +97,17 @@ describe("workspace protocol 'workspace:' specifiers", () => {
       expect((writePkg as any).updatedManifest('package-2').dependencies).toMatchObject({
         'package-1': '1.1.0', // workspace:*
       });
+      expect((writePkg as any).updatedManifest('package-2').peerDependencies).toMatchObject({
+        // peer deps should be bumped to a fixed minor version
+        'package-1': '1.1.0', // workspace:^
```
ghiscoding, line 102: the comment is wrong, it should be `workspace:*` instead of `workspace:^`

ghiscoding:
The `workspace:` protocol documentation is separate and can be found under both [version#workspace-protocol](https://github.com/lerna-lite/lerna-lite/tree/main/packages/version#workspace-protocol) and [publish#workspace-protocol](https://github.com/lerna-lite/lerna-lite/tree/main/packages/publish#workspace-protocol) which is literally a copy of pnpm [`workspace:`](https://pnpm.io/workspaces#publishing-workspace-packages). I don't really want to add more docs than what I have now for peer deps, since `workspace:` docs already covers that. in pnpm `workspace:*` resolves to the exact version (ie `1.2.0`) and looking at Yarn [`workspace:`](https://yarnpkg.com/features/workspaces#publishing-workspaces) documentation, it is also doing the same thing and using `workspace:*` in a peer dependencies will also resolve to the current exact version (I found a comment to be wrong in the PR which is related to `workspace:*` and its output is correct)

ma-multipla:
> The `workspace:` protocol documentation is separate and can be found under both [version#workspace-protocol](https://github.com/lerna-lite/lerna-lite/tree/main/packages/version#workspace-protocol) and [publish#workspace-protocol](https://github.com/lerna-lite/lerna-lite/tree/main/packages/publish#workspace-protocol) which is literally a copy of pnpm [`workspace:`](https://pnpm.io/workspaces#publishing-workspace-packages). I don't really want to add more docs than what I have now for peer deps (maybe just a link to point to these other docs), since `workspace:` docs already covers that info. in pnpm `workspace:*` resolves to the exact version (ie `1.2.0`) and looking at Yarn [`workspace:`](https://yarnpkg.com/features/workspaces#publishing-workspaces) documentation, it is also doing the same thing and using `workspace:*` in a peer dependencies will also resolve to the current exact version (I found a comment to be wrong in my PR which is related to `workspace:*` and its output is correct)

I see. I just checked again, made a mistake reading the PR before. Looks good.

ghiscoding:
@ma-multipla fixed under the new [v2.2.0](https://github.com/lerna-lite/lerna-lite/releases/tag/v2.2.0) release. Thanks for the feedback and contributions

PR #687, opened 2023-08-07
bug: exit with code `1` on missing package installation

Naymi:
## Description

added rethrow err on miss package

## Motivation and Context

https://github.com/lerna-lite/lerna-lite/issues/686

## How Has This Been Tested?

test updated

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:


packages/cli/src/cli-commands/__tests__/cli-changed-commands.spec.ts
```diff
@@ -7,7 +7,9 @@ describe('ChangedCommand CLI options', () => {
   it('should log a console error when ChangedCommand is not provided', async () => {
     const consoleSpy = vi.spyOn(console, 'error').mockImplementation(() => {});
 
-    await cliChanged.handler(undefined as any);
+    await expect(cliChanged.handler(undefined as any)).rejects.toMatchInlineSnapshot(
+      '[Error: [vitest] vi.mock("@lerna-lite/changed", factory?: () => unknown) is not returning an object. Did you mean to return an object with a "default" key?]'
```
ghiscoding, line 11: that's not really the same as what we're intending to test, is that the best we can do to test the missing installation? Why are we getting this error instead of the original error?

Naymi: https://github.com/lerna-lite/lerna-lite/pull/687/files/7b745a4da3c90a9f8354e97ef61a89b83c8fc4a1#diff-588f81d55155499cea386ba856cfdae1dda3317889d707bab7374ab80a690418R3

ghiscoding: but that still wouldn't test the missing installation error. What do you see now when running the same `lerna run` command in your console? I expect that you now see the same error that is thrown, right?

Naymi: just added this in console
image

Naymi: > sure but it's nice to match the error in the unit tests as well when possible, if it's too hard then we can disregard

the difficulty is that I did not find a way to indicate to vitest the dependency as not available, only use the `mock` format
In theory one could:
- remove/rename link to `node_modules` before this test. The decision looks very controversial, but honest
- throw new error with text from `console.error` on cli-version-commands.ts:313

Naymi: maybe so?
```

  handler: async (argv: VersionCommandOption) => {
    try {
      // @ts-ignore
      // eslint-disable-next-line
      const { VersionCommand } = await import('@lerna-lite/version');
      new VersionCommand(argv);
    } catch (err: unknown) {
      throw new Error(`"@lerna-lite/version" is optional and was not found. Please install it with "npm install @lerna-lite/version -D". ${err}`)
    }
  },
```

Naymi:
Check please)

PR #815, opened 2024-02-09
fix(list): explicitly exit upon completion

ghiscoding:
## Description

As per original Lerna [commit](https://github.com/lerna/lerna/commit/bafe090e837675c0febb938bab3e8c71b27dbe70)

## Types of changes

## Checklist:


ghiscoding:
as per all failing unit tests, this approach doesn't seem like a good one, so let's cancel this PR

PR #844, opened 2024-04-15
fix(publish): downgrade npm-packlist to v5.1.1

wassim-ben-amor:
## Description
Downgrade `npm-packlist` dependency to v5.1.3

## Motivation and Context
On `npm-packlist@` [v6.0.0](https://github.com/npm/npm-packlist/releases/tag/v6.0.0), exactly on [c37371b](https://github.com/npm/npm-packlist/commit/c37371b060ee42ebaebf134f37471c8330fc4d27#diff-92bbac9a308cd5fcf9db165841f2d90ce981baddcb2b1e26cfff170929af3bd1R208-R211) a breaking change is introduced.

The generation of the tree using:
```
const arborist = new Arborist({ path: this.path })
arborist.loadActual().then((tree) => {
```
is a very long process when it's done on monorepos that uses `pnpm` and `^workspace` protocol which may causes CI timeouts when publishing many packages at the same time. The slowness is explained by running the [findMissingEdges](https://github.com/npm/cli/blob/latest/workspaces/arborist/lib/arborist/load-actual.js#L189-L191) part on `Arborist`. It can be avoided by simply doing:
`const tree = await arborist.loadActual({ignoreMissing: true});`
But we are not totally sure about its impact. That's why I opened a https://github.com/npm/npm-packlist/issues/225 to understand the impact.

This dependency is not bumped on [lerna repo](https://github.com/lerna/lerna/blob/main/packages/lerna/package.json#L82). In this PR, a downgrade to `v5.1.3` is done to follow the [same usage of it on lerna](https://github.com/lerna/lerna/blob/main/libs/core/src/lib/pack-directory.ts#L79).
It will be better to wait for its bump on `lerna` before bumping it on `lerna-lite` .

The issue is more described in this [thread](https://github.com/lerna-lite/lerna-lite/discussions/843).

## How Has This Been Tested?
By just patching lerna-lite in my monorepo with the same changes. It do the publish instantly.

## Types of changes

## Checklist:

- [] My change requires a change to the documentation.
- [] I have updated the documentation accordingly.
- [] I have added tests to cover my changes.


ghiscoding:
since you mentioned that v7 is the breaking change, I wonder if we could use v6 instead of v5? I'm just trying to get newer versions has much as possible to avoid security issues popping up

wassim-ben-amor:
Hello @ghiscoding , thank you for checking.
After testing the v6.0.1, I have found that I have the same slowness issue 😞 . After checking the changelog, it appears that arborist logic was added exactly on v6 [here](https://github.com/npm/npm-packlist/commit/c37371b060ee42ebaebf134f37471c8330fc4d27#diff-92bbac9a308cd5fcf9db165841f2d90ce981baddcb2b1e26cfff170929af3bd1R211-R214).
On v7 they removed the Arborist dependency and made it as a parameter only https://github.com/npm/npm-packlist/commit/123875a9db0d18c2ba91e9841ab8e15df3057e18.

Sadly, it's working fine only with v5.1.3 at maximum. I have updated the PR and the description if you are ok with it.

ghiscoding:
alright in that case it seems that we are stuck with that version then, let's hope that they reply to your question and that they provide a way to disable the check in the future. So let's hope that they address sometime in the future. I might reach out to you again in the future to test newer version if they make any change, since I can't replicate this slowness on my side. Thanks, I'll push a newer version sometime this week

wassim-ben-amor:
Thank you,
I will be glad to continue on the subject sure !
Just a notice, npm-packlist already provide a way to ignore that slow step with using **ignoreMissing**:
```
const tree = await arborist.loadActual({ignoreMissing: true});
const files: string[] = await packlist(tree);
```
I think it's just a matter of confirmation from their side that it has no side effects on the expected output.

ghiscoding:
ok I see, I'd be happy for sure to receive another PR from your side if that turns out to be true. Cheers and thanks for the contribution

wassim-ben-amor:
Yes, I already tried it and it works perfectly without slowness.

ghiscoding:
@wassim-ben-amor just so you know, Renovate keeps finding dependencies to update, I'm hoping that this won't bring back the slowness that you had before but just in case, please report back in the project is this slowness comes back for you (for example [v3.4.0](https://github.com/lerna-lite/lerna-lite/releases/tag/v3.4.0) updated to a new major of Pacote and some other npm registry deps).

ghiscoding:
@wassim-ben-amor so this change with latest `npm-packlist` is now included in the latest Lerna release, they have done the exact same change that I've done (in this [commit](https://github.com/lerna/lerna/pull/4033/commits/b47c9fa620bb7c92d7bec33d9f991cd646396624#diff-9d7bc19db3fc00274f88785a57f8e521252f7dbeaae23d1bb217848f7229fd27)) and I really want upgrade back to the latest version of `npm-packlist` since I don't want to stay on old version forever and it doesn't look like anyone replied to your question on their repo... If you think that the `loadActual({ ignoreMissing })` can be a good alternative, then I could add an option like `--arborist-ignore-missing` that you could enable yourself. I'd really to upgrade back to the latest release as much as possible.

ghiscoding:
@wassim-ben-amor I'm reimplementing the latest version of `npm-packlist` (because I don't want to stay on older version of this dependency), however I'm adding an optional `arboristLoadOptions` that will allow the user (you) to provide options like `{ ignoreMissing: true }` to the `arborist.loadActual(options)` method. See new PR #889

PR #871, opened 2024-05-31
feat(version): add `--push-tags-one-by-one`, fixes #701

### LINKED ISSUE #701: Allow to push tags one by one to overcome GitHub limit
### Clear and concise description of the problem

Hi, I run into a weird limitation. Github workflows aren't triggered when more than 3 tags are pushed simultaneously. This becomes relevant in `independent` mode. See here for more context https://github.com/lerna/lerna/issues/2218. I first wanted to use `--no-follow-tags` but it doesn't exist in lerna-lite. `--no-push` also has no impact on pushing tags. https://github.com/lerna-lite/lerna-lite/blob/3e7dcb7fcc9bb5d56b57141f23572a67b8e3b073/packages/version/src/version-command.ts#L840

I tested https://github.com/lerna/lerna/issues/2218#issuecomment-1662185323 which unfortunately doesn't work :disappointed:

### Suggested solution

Introduce a new field `--push-tags-serial` to prevent pushing all tags simultaneously.

### Alternative

_No response_

### Additional context

_No response_

### Validations

- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

ghiscoding:
## Description

Push Git tags one by one

## Motivation and Context

- possibly fixes #701 
- when running Lerna version in `independent` mode, user could overcome GitHub limit

## Types of changes

## Checklist:

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,6 +1,6 @@
 {
   "$schema": "node_modules/@lerna-lite/cli/schemas/lerna-schema.json",
-  "version": "3.4.0",
+  "version": "independent",
   "loglevel": "verbose",
   "npmClient": "pnpm",
   "command": {
@@ -16,7 +16,8 @@
       "createRelease": "github",
       "changelogIncludeCommitsClientLogin": " - by @%l",
       "changelogHeaderMessage": "## Automate your Workspace Versioning, Publishing & Changelogs with [Lerna-Lite](https://github.com/lerna-lite/lerna-lite) 📦🚀",
-      "message": "chore(release): publish new version %v"
+      "message": "chore(release): publish new version %v",
+      "pushTagsOneByOne": true
     }
   },
   "changelogPreset": "conventional-changelog-conventionalcommits",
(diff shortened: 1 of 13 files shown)


ghiscoding:
@StarpTech I'm not at all sure if that would fix the issue you brought in issue #701, so I'd like to have your opinion. I tried in dry-run mode by switching Lerna-Lite from fixed version to independent mode. I also did not try the Git commands either, so I have no idea if it would work or fail.

So do you think that would work or is it going in the wrong direction? Because if that doesn't work then I have no idea on how else to address the problem. 

Here's 1 Git command
```sh
git push --no-verify --atomic origin @lerna-lite/cli@4.0.0 main
```

image

ghiscoding:
no feedback provided, but let's go ahead with it in a new release. @StarpTech if you have a chance to try it in the future and confirm that it works, that would be great. I think the PR will fix the GitHub limit. Cheers

StarpTech:
Hi, in your repo, were you able to reproduce the issue?

ghiscoding:
@StarpTech no since the simple repo I created to test this only has 2 packages so I won't hit the limit, however I tested the git push 1 by 1 and it seems to be working as expected.

PR #897, opened 2024-07-14
fix(core): use inquirer new modern approach to move away from legacy

ghiscoding:
## Description

Use the new `inquirer` modern approach which is ESM

## Motivation and Context

Since `inquirer` v10, the previous approach is now tagged as being the "legacy" approach and I'm assuming that might be removed in the future (or at least, less supported), so let's migrate to the new approach. Also since the new approach is ESM and targeted to only import what we use, that should decrease our build size even if that might be a slight decrease. 

As mentioned on the `inquirer` GitHub [page](https://github.com/SBoudrias/Inquirer.js/tree/main/packages/inquirer)

> This is the legacy version of Inquirer.js. While it still receives maintenance, it is not actively developed. For the new Inquirer, see [@inquirer/prompts](https://www.npmjs.com/package/@inquirer/prompts).

## Types of changes

## Checklist:

### DIFF
--- a/packages/core/src/prompt.ts
+++ b/packages/core/src/prompt.ts
@@ -64,7 +64,14 @@ export async function promptSelectOne(
  * @param {Omit<Question, 'choices'>} [options]
  * @returns {Promise<string>}
  */
-export async function promptTextInput(message: string, { filter, validate }: Omit<Question, 'choices'> = {}): Promise<string> {
+type InputConfig = Parameters<typeof input>[0];
+export async function promptTextInput(
+  message: string,
+  {
+    filter,
+    validate,
+  }: Pick<InputConfig, "validate"> & { filter?: (value: string) => string }
+): Promise<string> {
   log.pause();
 
   const promptResult = await input({
--- a/packages/core/src/prompt.ts
+++ b/packages/core/src/prompt.ts
@@ -1,4 +1,6 @@
-import inquirer from 'inquirer';
+import expand from '@inquirer/expand';
+import input from '@inquirer/input';
+import select from '@inquirer/select';
 import { log } from '@lerna-lite/npmlog';
 
 interface ListChoiceOptions {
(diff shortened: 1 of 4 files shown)


packages/core/src/prompt.ts
```diff
@@ -21,67 +23,54 @@ interface Question {
...
 
 /**
  * Prompt for selection
  * @param {string} message
- * @param {{ choices: import("inquirer").ListChoiceOptions[] } & Pick<import("inquirer").Question, 'filter' | 'validate'>} [options]
+ * @param {Question} [options]
  * @returns {Promise<string>}
  */
-export async function promptSelectOne(message: string, { choices, filter, validate } = {} as Question): Promise<string> {
+export async function promptSelectOne(message: string, { choices }: Question = {} as Question): Promise<string> {
```
SBoudrias, line 45:
```suggestion
export async function promptSelectOne(
  message: string,
  { choices }: Partial<Parameters<typeof select>[0]> = {}
): Promise<string[]> {
```

With `as` you're ignoring type check, and you can see if breaks the typing since it allow you to return `Promise<string>` while `select` returns an array, not a string 😬

packages/core/src/prompt.ts
```diff
@@ -21,67 +23,54 @@ interface Question {
...
 
 /**
  * Prompt for input
  * @param {string} message
- * @param {Pick<import("inquirer").Question, 'filter' | 'validate'>} [options]
+ * @param {Omit<Question, 'choices'>} [options]
  * @returns {Promise<string>}
  */
-export async function promptTextInput(message: string, { filter, validate } = {} as Omit<Question, 'choices'>): Promise<string> {
+export async function promptTextInput(message: string, { filter, validate }: Omit<Question, 'choices'> = {}): Promise<string> {
```
SBoudrias, line 64:
```suggestion
type InputConfig = Parameters<typeof input>[0];
export async function promptTextInput(
  message: string,
  {
    filter,
    validate,
  }: Pick<InputConfig, "validate"> & { filter?: (value: string) => string }
): Promise<string> {
```

ghiscoding:
@SBoudrias thanks a lot for the review, it was quite helpful and if you're looking for a lighter version of Lerna, then please consider Lerna-Lite 😉 

Merci, bonsoir 🌙

PR #925, opened 2024-09-06
fix(version): Renamed 'skip-bump-only-release' CLI argument to 'skip-…

### LINKED ISSUE #924: 'version' command | Unknown argument "skipBumpOnlyReleases"
### Describe the bug

When using lerna-lite version argument '--skip-bump-only-releases' https://github.com/lerna-lite/lerna-lite/blob/main/packages/version/README.md#--skip-bump-only-releases, lerna-lite version does not execute, but fails with:

    ERR! lerna-lite Unknown arguments: skip-bump-only-releases, skipBumpOnlyReleases

Using '--skip-bump-only-release' works fine, though.

Related issue #703 => pr #721
    

### Expectation

lerna-lite version executes fine with '--skip-bump-only-releases'.

### Reproduction

    $ npx -y -p conventional-changelog-conventionalcommits@7 -p @lerna-lite/cli@3.9.0 -p @lerna-lite/version@3.9.0 \
      lerna version --yes --skip-bump-only-releases --dry-run

    ERR! lerna-lite Unknown arguments: skip-bump-only-releases, skipBumpOnlyReleases

### Lerna config and logs

_No response_

### Environment Info

$ npm -v
10.8.1

### Used Package Manager

npm

### Validations

- [X] Check that there isn't [already an issue](https://github.com/lerna-lite/lerna-lite/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/lerna-lite/lerna-lite/discussions).

SchulteMarkus:
## Description

Renamed 'skip-bump-only-release' to 'skip-bump-only-releases' for consistency. 
Added alias 'skip-bump-only-release' for backwards compatibility and refined the description for clarity. 

## Motivation and Context

Fixes #924

## How Has This Been Tested?

Just CLI argument definition

## Types of changes

## Checklist:

### DIFF
--- a/packages/cli/src/cli-commands/cli-version-commands.ts
+++ b/packages/cli/src/cli-commands/cli-version-commands.ts
@@ -281,8 +281,9 @@ export default {
         describe: 'Runs `npm install --package-lock-only` or equivalent depending on the package manager defined in `npmClient`.',
         type: 'boolean',
       },
-      'skip-bump-only-release': {
-        describe: 'do we want to skip creating a release (github/gitlab) when the version is a "version bump only"?',
+      'skip-bump-only-releases': {
+        describe: 'Skip release(s) creation when the version is a "version bump only"?',
+        alias: 'skip-bump-only-release',
         type: 'boolean',
       },
       'premajor-version-bump': {


packages/cli/src/cli-commands/cli-version-commands.ts
```diff
@@ -281,8 +281,9 @@ export default {
         describe: 'Runs `npm install --package-lock-only` or equivalent depending on the package manager defined in `npmClient`.',
         type: 'boolean',
       },
-      'skip-bump-only-release': {
-        describe: 'do we want to skip creating a release (github/gitlab) when the version is a "version bump only"?',
+      'skip-bump-only-releases': {
+        describe: 'Skip release(s) creation when the version is a "version bump only"?',
+        alias: 'skip-bump-only-release',
```
ghiscoding, line 286: is this really needed? I don't think that the singular option will do anything because the option in VersionCommand is really plural not singular, so I'm assuming that this alias will do.. just nothing

https://github.com/lerna-lite/lerna-lite/blob/8d028847e8237a9b8ab94e36b23050aae219629c/packages/version/src/version-command.ts#L410-L413

SchulteMarkus: I removed the alias.
I think this is a breaking change. Users who previously used the singular argument will now receive an error message. But this may be even better than the current state (command executes, but argument has no effect).
In the Conventional Commit sense, however, this change should then lead to a major release, I think.

ghiscoding:
Thanks for the contribution :)

ghiscoding:
published as v3.9.1, thanks again 🎉

PR #932, opened 2024-09-16
fix(version): Update default "describe [Git] tag" pattern in non-independent mode to 'v*'

SchulteMarkus:
## Description

Changed the default --describe-tag pattern from an empty string to 'v*' for non-independent mode in the `lerna version` command. Updated related documentation and tests to reflect this change. 
This change is being done to have --describe-tag and --tag-version-prefix in sync for non-independent mode.

## Motivation and Context

When using `fixed` mode, I ran into similar problems as described in #220. Setting `describeTag` to `v*` helped.

In my opinion, it should be the default to have `describeTag` and `tagVersionPrefix` synced.

## How Has This Been Tested?

I wasn't able to test locally.

## Types of changes

## Checklist:

### DIFF
--- a/packages/core/src/utils/collect-updates/__tests__/collect-updates.spec.ts
+++ b/packages/core/src/utils/collect-updates/__tests__/collect-updates.spec.ts
@@ -507,7 +507,7 @@ describe('collectUpdates()', () => {
     expect(describeRefSync).toHaveBeenCalledWith({ cwd: '/test', match: '*custom-tag*' }, true);
   });
 
-  it('no use "describeTag" in non-independent mode', async () => {
+  it('use "describeTag" with default value in non-independent mode', async () => {
     const graph = buildGraph();
     const pkgs = graph.rawPackageList;
     const execOpts = { cwd: '/test' };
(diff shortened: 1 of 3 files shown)


ghiscoding:
Taking a look at the original PR which brought this describe, it was already discussed to not use `v*` as the default and you're trying to add it back. Even if it's acceptable, this seems like a breaking change and as mentioned in the discussion, you can change the describe prefix with [`--tag-version-prefix`](https://github.com/lerna-lite/lerna-lite/tree/main/packages/version#--tag-version-prefix)

SchulteMarkus:
Yes, I am using a configuration in which Lerna-Lite is not be distracted by non-Lerna Git tags (#220). 
I just find it illogical that (in fixed mode) this is the normal case. But this decision is not up to me.

ghiscoding:
I searched at the history to see when this tag pattern was introduced because I wanted to know if it was the default in Lerna and it's not, it was actually introduced in Lerna-Lite via this [change](https://github.com/lerna-lite/lerna-lite/pull/267/files#diff-37f9f605e9e81a676b7e0f21e52868f3e177a0eb0857cb89e73e7f7e83443213), I know that the user who introduced this PR is typically using independent mode, so he could have been wrong for the fixed mode. 

So I would say that we could potentially merge your PR in the future but like I said this is a breaking change and I don't plan to do any breaking change until the next NodeJS version is EOL which is only in Spring of next year. So the best I could do is to add it to my roadmap of [Future Changes](https://github.com/lerna-lite/lerna-lite/discussions/763) and change this default on the next major in about 6 months

ghiscoding:
@SchulteDev 
Prepping for upcoming major version, it's now time to reopen the PR :)

I wanted to reopen your PR and keep your contribution but unfortunately the branch was deleted, so I went ahead and replicated the changes in PR #1001

ghiscoding:
this is now shipped, see the new major [v4.0](https://github.com/lerna-lite/lerna-lite/releases/tag/v4.0.0) release

PR #948, opened 2024-10-12
feat: support --scope filter options in all lerna commands

farfromrefug:
…ersioning)

I ran some basic tests and it seemed to work prefectly fine.


ghiscoding:
The build is failing because your change causes circular dependencies, `filter-package` is calling `core` and `core` is calling `filter-package`, this won't work. 

https://github.com/lerna-lite/lerna-lite/blob/37f5f9758417b779d2f8636949364df35be87305/packages/filter-packages/package.json#L37-L39

The main reason to create it as a separate package was because the core wasn't using it and not all commands needed the filtering scope, so I moved it into its own package. To remove the circular dependencies issue, you'll probably have to move all the filtering stuff to the core package but you're now also increasing the build size for everyone, I'm not sure that everyone would want it though. 

Also it would probably require documentation updates and most certainly some unit tests

basic scope test for `lerna run` is this one below and I would expect to see tests for version/publish most probably

https://github.com/lerna-lite/lerna-lite/blob/37f5f9758417b779d2f8636949364df35be87305/packages/run/src/__tests__/run-command.spec.ts#L159-L163

the `lerna publish` command unit test also has this test which would become invalid with your PR, that would have to be addressed too

https://github.com/lerna-lite/lerna-lite/blob/37f5f9758417b779d2f8636949364df35be87305/packages/publish/src/__tests__/publish-command.spec.ts#L122-L131

farfromrefug:
@ghiscoding ok sure i am ok refactoring. About the size increase i dont think that will be much. the filter-package is quite small. 
Let me know how you want me to proceed

packages/core/src/command.ts
```diff
@@ -316,10 +318,16 @@ export class Command<T extends AvailableCommandOption> {
     let chain: Promise<any> = Promise.resolve();
 
     if (this.commandName !== 'info') {
+      this.options.isIndependent = this.project.isIndependent();
       chain = chain.then(() => this.project.getPackages());
       chain = chain.then((packages) => {
         const { graphType } = this.options.command?.[this.commandName] ?? {};
         this.packageGraph = new PackageGraph(packages || [], graphType ?? 'allDependencies');
+        if (this.options.isIndependent) {
```
ghiscoding, line 326: shouldn't that also be working with fixed mode and not just independent?

ghiscoding:
> @ghiscoding ok sure i am ok refactoring. About the size increase i dont think that will be much. the filter-package is quite small. Let me know how you want me to proceed

yeah looking at the size, it seems acceptable, however as I mentioned above I would like to this that this might also work in fixed mode as well

packages/core/src/command.ts
farfromrefug: @ghiscoding yes i can make it work in both cases. But does it really makes sense? Because if you use version with scope in fixed mode then you end up with a package with a different version than others.
This is the only reason why i added that if case

ghiscoding: I think it does but in very rare occasion like maybe patch versions or other things that I can't think of

farfromrefug:
@ghiscoding so i did refactor and put `filter-packages` in the core. I also modified the tests for publish. I am not sure though what to be "expected" from the test output.
i ll add one for version once we define what output we want

ghiscoding:
> @ghiscoding so i did refactor and put `filter-packages` in the core. I also modified the tests for publish. I am not sure though what to be "expected" from the test output. i ll add one for version once we define what output we want

Yeah I'm not sure either how to test these, I would say to look at the existing tests and do something similar. On that end, there are still a few tests failing, 1 of them seems like a copy+paste error because "profilter-packagest " surely doesn't exist

> ❯ packages/list/src/__tests__/list-command.spec.ts:17:25
> Caused by: Caused by: Error: Failed to load url ../../../core/src/profilter-packagest (resolved id: ../../../core/src/profilter-packagest). Does the file exist?

farfromrefug:
@ghiscoding  i fixed what i could. Now i am not sure why the tests are still failing. Any idea?

ghiscoding:
> @ghiscoding i fixed what i could. Now i am not sure why the tests are still failing. Any idea?

possibly yes because it looks like you forgot to adjust the imports inside the `get-filtered-package.ts` (I'm not sure if there's more but possibly), you can't import from `@lerna-lite/core`, it needs to be local imports instead. 

https://github.com/lerna-lite/lerna-lite/blob/7503995d6330ca371d44a15d0eff4814accbbcf8/packages/core/src/filter-packages/lib/get-filtered-packages.ts#L1-L2

the mocks will also be adjusted in consequence

https://github.com/lerna-lite/lerna-lite/blob/7503995d6330ca371d44a15d0eff4814accbbcf8/packages/core/src/filter-packages/__tests__/get-filtered-packages.spec.ts#L5-L8

ghiscoding:
I got the `filtered-package.spec.ts` working after doing these changes, so if you do these changes it works

```ts
vi.mock('../../prompt', async () => await vi.importActual('../../__mocks__/prompt'));
vi.mock('../../utils/output', async () => await vi.importActual('../../__mocks__/output'));
vi.mock('../../utils/check-working-tree', async () => await vi.importActual('../../__mocks__/check-working-tree'));
vi.mock('../../utils/collect-updates/collect-updates.js', async () => await vi.importActual('../../__mocks__/collect-updates'));
```
and then couple lines below, note that the "mocked modules" is the module itself being imported and mocked **not** a path to the mock. 

```ts
// mocked modules
import { collectUpdates } from '../../utils/collect-updates/collect-updates.js';
```

farfromrefug:
@ghiscoding thanks a lot you did all the heavy lifting!

packages/list/src/__tests__/list-command.spec.ts
```diff
@@ -12,14 +12,16 @@ vi.mock('@lerna-lite/core', async () => ({
...
+  getFilteredPackages: (await vi.importActual<any>('../../../core/src/filter-packages')).getFilteredPackages,
 }));
 
 vi.mock('@lerna-lite/list', async () => await vi.importActual<any>('../list-command'));
-vi.mock('@lerna-lite/filter-packages', async () => await vi.importActual<any>('../../../filter-packages/src/get-filtered-packages'));
 
 // mocked modules
 import { collectUpdates, ListCommandOption, logOutput } from '@lerna-lite/core';
 
+vi.mock('../../../core/src/utils/collect-updates/collect-updates.js', async () => await vi.importActual('../../../core/src/__mocks__/collect-updates'));
```
ghiscoding, line 23: could you move this mock with the other mocks above? around line 17 would be better

packages/core/src/command.ts
```diff
@@ -316,10 +318,14 @@ export class Command<T extends AvailableCommandOption> {
     let chain: Promise<any> = Promise.resolve();
 
     if (this.commandName !== 'info') {
+      this.options.isIndependent = this.project.isIndependent();
```
ghiscoding, line 321: is this necessary?

farfromrefug: should not be no

farfromrefug:
@ghiscoding I think i addressed all your comments

packages/version/src/__tests__/version-bump.spec.ts
```diff
@@ -124,6 +124,26 @@ describe('version bump', () => {
...
   });
 
+  test('prerelease increments version for the filtered packages when providing a --scope', async () => {
+    const testDir = await initFixture('independent');
+
+    // await new VersionCommand(createArgv(testDir, "--bump", "prerelease"));
+    await factory(createArgv(testDir, '--bump', 'prerelease', '--scope', 'package-1'));
+
+    const message = await getCommitMessage(testDir);
+    expect(message).toContain('package-1@1.0.1-alpha.0');
```
ghiscoding, line 134: the `message` here is different compare to the Publish tests, this `message` is a string (or a string array, I forgot) that will contain (or not) all the packages. So yes you did check the pkg-1 **but** you should also check that the other packages are **not** included in that `message`, that is basically what I wrote and meant in my previous comment

> and then the last expect should be an inverse to make sure it's filtered out after the scope
> 
> ```ts
>  expect(message).not.toContain('package-5@5.0.1-alpha.0');
> ```

farfromrefug: @ghiscoding sorry i missed that

farfromrefug: btw i think there is an issue with those tests. If i did the ignore test with package 5 then the toContain was not working. Feels like the `message` passed to `expect` is truncated or something. I had to use package 1 or i could not get it to work

ghiscoding: you can add a breakpoint to your test and see the content of `message` then adjust your test according to it

packages/version/src/__tests__/version-bump.spec.ts
```diff
@@ -131,7 +131,11 @@ describe('version bump', () => {
     await factory(createArgv(testDir, '--bump', 'prerelease', '--scope', 'package-1'));
 
     const message = await getCommitMessage(testDir);
-    expect(message).toContain('package-1@1.0.1-alpha.0');
```
ghiscoding: ahh sorry could you also add this one back? I meant to have both the filtered package and the ones that are outside of your scope. So `contain(pkg-1)` and all other ones as `not.contain(...)`

packages/version/src/__tests__/version-bump.spec.ts
farfromrefug: @ghiscoding i did the debugger was showing the full text with line breaks but the `.toContain` was not working

Like this is not working when i think it should:
```ts
test('prerelease increments version only for the packages that are not --ignore(d)', async () => {
    const testDir = await initFixture('independent');

    // await new VersionCommand(createArgv(testDir, "--bump", "prerelease"));
    await factory(createArgv(testDir, '--bump', 'prerelease', '--ignore', 'package-@(2|3|4|5|1)'));

    const message = await getCommitMessage(testDir);
    expect(message).toContain('package-6@1.0.1-alpha.0');
  });

```

image

farfromrefug: EDIT: my bad i see now that versions were not the same for all packages... i misread the version numbers!
All good

ghiscoding (approved):
looking good, thanks a lot for all your effort and patience :)

I'll merge and publish in the next couple day or maybe tonight, we'll see. Thanks again

farfromrefug:
@ghiscoding and thank you for taking the time to help me!

ghiscoding:
now available in [v3.10.0](https://github.com/lerna-lite/lerna-lite/releases/tag/v3.10.0), thanks again for the contribution :)

PR #965, opened 2024-12-26
feat(core): support catalog: protocol

Mister-Hope:
## Description

This PR support pnpm `catalog:` protocal.

## Motivation and Context

https://github.com/lerna-lite/lerna-lite/discussions/901
https://github.com/lerna-lite/lerna-lite/issues/902

## How Has This Been Tested?

Tested locally in vuepress/ecosytem by patching the core pacakge.

## Types of changes

## Checklist:


ghiscoding:
Thanks for the PR, the code looks ok but I can't really merge this without any unit tests in place. 
There's maybe 2 ways to add tests, it could be via 
- package-graph spec 
  - https://github.com/lerna-lite/lerna-lite/blob/main/packages/core/src/package-graph/__tests__/package-graph.spec.ts 
- or a separate spec file like this
  - https://github.com/lerna-lite/lerna-lite/blob/main/packages/publish/src/__tests__/publish-workspace-protocol-specifiers.spec.ts

ghiscoding:
I decided to go ahead and merge, even if we should really add unit tests eventually. We should be able to add tests in the future even without a global `pnpm-workspace.yaml` file, we should be able to mock for example the yaml file import with a custom one and test with that... anyway, let's merge and get feedback from users (if any) and let's get the ball rolling if that can unblock some users of catalog. Thanks

ghiscoding:
@Mister-Hope have you actually tried the new version? Because I tried it on a project of mine and I still see the `catalog:` in my final package. Perhaps it doesn't work in `--dry-run` mode which is what I tried it? I don't think there's any flag to enable for this to work right? I didn't test in production mode and I'm not sure that I want to test it either, I would think that `--dry-run` mode should work too

image

ghiscoding:
what do you mean by recognize? just not crash when publishing, is that what you mean? in your PR above you said it doesn't bump the version of catalog dependencies which I don't care, but if it doesn't even replace the `catalog:` with the associated ones in the `pnpm-workspace.yml` then it seems useless.

PR #1011, opened 2025-04-12
refactor: improve codes

Mister-Hope:
This PR removes some unnecessary checks/optional-chaining and wrong await


ghiscoding:
it seems ok, thanks for the contribution

PR #1012, opened 2025-04-12
fix(core): fix wrong regexp about workspace protocol

Mister-Hope:


Mister-Hope:
No actual problem  but `|` is matched previously and it's not in spec.

ghiscoding:
ah yeah ok, I used it as separator but you're right, there's no need. Thanks

PR #1066, opened 2025-06-12
Fix example of --dry-run in version README.md

theodorekouzelis:
The example given for the flag --dry-run was using the run/watch command rather than version

## Description

The example given for the flag --dry-run was using the run/watch command rather than version

## Motivation and Context

 I believe the documented command was wrong and correcting would reduce confusion in the future.

## How Has This Been Tested?

By running the documented example locally

## Types of changes

- [ x] Chore (change that has absolutely no effect on users)

## Checklist:

- [x ] My code follows the code style of this project.
- [ x] My change requires a change to the documentation.
- [ x] I have updated the documentation accordingly.
- [ x] I have added tests to cover my changes.
- [ x] All new and existing tests passed.


ghiscoding:
ahh thanks, another copy + paste typo 😆

PR #1104, opened 2025-08-08
feat: use native `structuredClone`

ghiscoding:
## Description

Drop another dependency and use native code

## Motivation and Context

[`structuredClone`](https://nodejs.org/api/globals.html#structuredclonevalue-options) is natively available as of Node v17, so no need to have an extra dependency for this anymore

## Types of changes

## Checklist:


ghiscoding:
let's close for now since doesn't seem to pass unit tests, may revisit in the future

PR #1105, opened 2025-08-08
refactor: use native object spreading

MichaelDeBoey:
## Motivation and Context

~~[`structuredClone`](https://nodejs.org/api/globals.html#structuredclonevalue-options) is natively available as of Node v17, so no need to have an extra dependency for this anymore~~

We could use object spreading since we're only accessing root properties

As per e18e initiative
https://github.com/e18e/ecosystem-issues/issues/187

## Types of changes

## Checklist:

---

I'm sure people from the wider @e18e ecosystem cleanup (like @43081j, @benmccann, @Fuzzyma, @outslept & @talentlessguy) will be very happy to see these kind of changes as well

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -42,7 +42,6 @@
     "@lerna-lite/npmlog": "workspace:*",
     "@npmcli/run-script": "^9.1.0",
     "ci-info": "^4.3.0",
-    "clone-deep": "^4.0.1",
     "config-chain": "^1.1.13",
     "dedent": "catalog:",
     "execa": "catalog:",
@@ -67,7 +66,6 @@
     "zeptomatch": "catalog:dependencies"
   },
   "devDependencies": {
-    "@types/clone-deep": "^4.0.4",
     "@types/dedent": "catalog:devDependencies",
     "@types/fs-extra": "catalog:devDependencies",
     "@types/glob-parent": "^5.1.3",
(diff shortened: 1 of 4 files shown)


ghiscoding:
Hey thanks for the PR but I already tried that yesterday in this closed PR #1104 and it looks like something isn't working as expected because the unit tests are timing out. I'm assuming that it's probably because `structuredClone` doesn't deep copy everything, for example I'm not sure if it copies functions properly? Anyway, I can't really go with this if the test are timing out. Thanks

outslept:
> Hey thanks for the PR but I already tried that yesterday in this closed PR #1104 and it looks like something isn't working as expected because the unit tests are timing out. I'm assuming that it's probably because `structuredClone` doesn't deep copy everything, for example I'm not sure if it copies functions properly? Anyway, I can't really go with this if the test are timing out. Thanks

The algo is described here: https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm

I believe if something is failing it is because of this - Function objects cannot be duplicated by the structured clone algorithm; attempting to throws a DataCloneError exception.

MichaelDeBoey:
@ghiscoding @outslept After some more looking at the code, I came to the conclusion that we can just use native object spreading, as we're only accessing root properties

ghiscoding:
Thank tests are all passing now and it looks like this change was also accepted in the original Lerna package. So I think it's safe to merge, thanks for the contribution 🚀

PR #1174, opened 2025-10-18
fix: Suppress ExitPromptError when leaving the prompt

heavenshell:
## Description

Hi,

https://github.com/SBoudrias/Inquirer.js/pull/1744 
This change prevents the ExitPromptError from being displayed when exiting the prompt with ctrl-c or ctrl-d,
which was noisy, so I made it so that the error is not displayed.

## Motivation and Context

When leaving lerna-lite prompot, SIGINT error has been shown.

Steps to reproduce
1. hit `lerna version --no-git-tag-version --no-push` command
2. A prompt to select the version is displayed
3. Hit Ctrl-C or Ctrl-D
4. ExitPromptError displayed

https://github.com/user-attachments/assets/8db728cb-4f65-447a-bbfc-dedfec4b7464

Solution:
Add ignore when the error is ExitPromptError.

https://github.com/user-attachments/assets/541d5ba0-d6f1-4e53-9830-da2f80796e44

## How Has This Been Tested?

I replaced the build-generated files with the test project and confirmed its behavior with the CLI (please let me know if there's a better way).

Sorry, I have no idea how to write test code for this.
I would be very grateful for any advice 🙏 

environment
macOS, VSCode terminal

## Types of changes

Sorry, I wasn't sure if this was a bug or a chore, but I marked it as a bug 🙏 

## Checklist:

### DIFF
--- a/packages/core/src/command.ts
+++ b/packages/core/src/command.ts
@@ -109,12 +109,13 @@ export class Command<T extends AvailableCommandOption> {
           // ValidationError does not trigger a log dump, nor do external package errors
           if (err.name !== 'ValidationError' && !err.pkg) {
             writeLogFile(this.project.rootPath);
-          }
 
-          if (err.name === 'ExitPromptError' && !err.pkg) {
-            // Suppress ExitPromptError when terminating the prompt
-            // (sending SIGINT)
-            return;
+            // Suppress ExitPromptError when terminating the prompt (sending SIGINT)
+            if (err.name === 'ExitPromptError') {
+              console.error('Termination call detected, exiting command');
+              resolve(null);
+              return;
+            }
           }
 
           warnIfHanging();


ghiscoding:
Hey thanks for the PR, however I'm not sure that just returning from the Promise is a good thing because that would leave it in a pending state I think and it also makes it hard to add unit tests. Also you can also group your code in the other `if` condition just above since it goes into it anyway. Would it work if we return and `resolve(null)`? Would that work?

```ts
// ValidationError does not trigger a log dump, nor do external package errors
if (err.name !== 'ValidationError' && !err.pkg) {
  writeLogFile(this.project.rootPath);

  // Suppress ExitPromptError when terminating the prompt (sending SIGINT)
  if (err.name === 'ExitPromptError') {
    console.error('Termination call detected, exiting command');
    resolve(null);
    return;
  }
}
```

with this, we can add a test

```ts
describe('with package error', () => {
 // ...

    it('does log when Prompt terminates', async () => {
      class ExitPromptError extends Error {
        override name = 'ExitPromptError';
      }

      class PkgErrorCommand extends Command<any> {
        initialize() {
          return true;
        }

        execute() {
          const err: any = new ExitPromptError('ExitPromptError');

          err.command = 'test-pkg-err';
          err.stdout = 'pkg-err-stdout';
          err.stderr = 'pkg-err-stderr';
          err.pkg = undefined;

          throw err;
        }
      }

      const command = new PkgErrorCommand({ cwd: testDir, stream: true } as any);

      await expect(command).resolves.toBeNull();
      expect(console.error).toHaveBeenCalled();
    });
  });
```

heavenshell:
Thank you for your review!
Exactly right.

Confirmed with resolve(null) works fine!
スクリーンショット 2025-10-19 12 27 37

I'll update patch and add test.
Thank you!

heavenshell:
@ghiscoding
Thank you for your feedback. I've fixed it accordingly.
https://github.com/lerna-lite/lerna-lite/pull/1174/commits/d03a718c92e73066c040a0bc21de3865c18f62a4 https://github.com/lerna-lite/lerna-lite/pull/1174/commits/91dd25dbda6b3525f867df30351190e7565df207

But the CI failed for a reason unrelated to this fix. Could you take a look? 🙏 

Thank you!

ghiscoding:
Yes there's a flaky test that exists since I updated Vitest, I haven't had a chance to look into that flaky test yet. The PR looks better, at least there's no pending promise and that's probably better since it won't leak memory.

Note that I will probably wait couple of weeks before releasing this because I just released a version couple days ago. Thanks for the contribution and you can upvote the project if you like it. Thanks again ⭐

heavenshell:
Thank you!

> I will probably wait couple of weeks before releasing this because I just released a version couple days ago

No problem! I'm happy to contribute this project! 😸 
(I forgot stared, but I just did!)

ghiscoding:
both of your PRs are now available in [v4.9.2](https://github.com/lerna-lite/lerna-lite/releases/tag/v4.9.2), thanks again for your contributions. :)

heavenshell:
Thank you for releasing!

PR #1176, opened 2025-10-23
fix: supress log file generation when terminating the prompt

heavenshell:
## Description

HI,

ref: https://github.com/lerna-lite/lerna-lite/pull/1174

This PR prevented errors from being displayed when exiting the prompt with Ctrl-C/Ctrl-D, 
but lerna-debug.log were still generated.
(I'm sorry I did not notice this at that time)

If the prompt is terminated using Ctrl-C or Ctrl-D, it is considered a normal exit, 
and lerna-debug.log will not be generated.

## Motivation and Context

Steps to reproduce

- hit `lerna version --no-git-tag-version --no-push` command
- A prompt to select the version is displayed
- Hit Ctrl-C or Ctrl-D, `Termination call detected, exiting command` is displayed
- hit `ls`, lerna-debug.log is generated

https://github.com/user-attachments/assets/80cc4cc5-85b3-40f1-aecb-8dbef94ee7db

Solution
Move `writeLogFile(this.project.rootPath);` after the ExitPromptError check.

https://github.com/user-attachments/assets/d1067d7d-48b8-4852-a76e-80e3a92b4996

environment
macOS, VSCode terminal

## How Has This Been Tested?

I replaced the build-generated files with the test project and confirmed its behavior with the CLI 

## Types of changes

## Checklist:

### DIFF
--- a/packages/core/src/command.ts
+++ b/packages/core/src/command.ts
@@ -108,14 +108,13 @@ export class Command<T extends AvailableCommandOption> {
 
           // ValidationError does not trigger a log dump, nor do external package errors
           if (err.name !== 'ValidationError' && !err.pkg) {
-            writeLogFile(this.project.rootPath);
-
             // Suppress ExitPromptError when terminating the prompt (sending SIGINT)
             if (err.name === 'ExitPromptError') {
               console.error('Termination call detected, exiting command');
               resolve(null);
               return;
             }
+            writeLogFile(this.project.rootPath);
           }
 
           warnIfHanging();


ghiscoding:
oh that is a really great catch, I also wouldn't have thought of looking at the debug logs 😄 
Also note that since it's low priority, I will probably wait a few more weeks before releasing a new version. 

Thanks a lot for the contribution 🙏🏻

heavenshell:
Thank you!

PR #1216, opened 2025-12-15
fix: use args to run `git show` command

ghoullier:
## Description

Hello 👋 
This pull request change the way `diffWorkspaceCatalog` execute git command under the hood.

When you call `execaSync` from `execa` you have to pass the process with args instead of the command including args.

The original error is catch by this [line](https://github.com/lerna-lite/lerna-lite/blob/77c98044dd761755f92b33a818abad6558991251/packages/core/src/utils/collect-updates/lib/make-diff-predicate.ts#L196)

## Motivation and Context

Without this change `lerna-lite` will not detect changes if you update the `pnpm-workspace.yaml`

## How Has This Been Tested?

I created a pnpm patch to test the fix see https://github.com/ghoullier/lerna-lite-pnpm-catalog-issue

## Types of changes

## Checklist:


ghiscoding:
I can merge but technically speaking it shouldn't make a difference since the latest execa version also accepts to run as a single string, so both approach of passing options (single string, array of strings) should work

PR #1358, opened 2026-07-02
fix(version): honor 0.x major-to-minor in prerelease bumps

ghiscoding:
## Description

Adjusted conventional version recommendation so 0.x major handling is applied before prerelease branching. This ensures breaking changes on 0.x do not escalate to 1.0.0 during prerelease recommendation paths.

Also added tests to cover the prerelease behavior for major-zero packages in both default and conventional bump prerelease modes.

## Motivation and Context

> **Note** this bug fix was originally created as a patch in the Agoric SDK repo via this [commit](https://github.com/Agoric/agoric-sdk/commit/ae1039bd40e6c50cbf8626e4d9a20220a64b11a1), so I'm just replicating it in Lerna-Lite

A logic gap allowed prerelease flows to bypass the 0.x compatibility rule that maps major to minor. That could produce premajor bumps from 0.x prereleases, which breaks expected zero-major compatibility behavior.

This change keeps prerelease and non-prerelease paths consistent for 0.x packages.

## How Has This Been Tested?

Ran targeted tests for conventional commits recommendation behavior:

- pnpm vitest run conventional-commits.spec.ts --pool forks --no-watch --config vitest.config.ts

Result:
- Exit Code: 0

## Types of changes

## Checklist:

### DIFF
--- a/packages/version/src/conventional-commits/__tests__/conventional-commits.spec.ts
+++ b/packages/version/src/conventional-commits/__tests__/conventional-commits.spec.ts
@@ -348,6 +348,43 @@ describe('conventional-commits', () => {
         const bump = await recommendVersion(pkg0, 'independent', {});
         expect(bump).toBe('0.2.0');
       });
+
+      it('keeps breaking changes in prerelease series by default when prereleaseId is set', async () => {
+        const cwd = await initFixture('major-zero');
+        const [pkg0] = await Project.getPackages(cwd);
+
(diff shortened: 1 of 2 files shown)


ghiscoding:
@turadg that PR was for you, which is basically your current patch that you have in Agoric SDK

michaelfig:
Thanks for your concerted effort to upstream our (Agoric's) patch and inform us of the update.  We'll definitely adopt it!

PR #1316, opened 2026-04-21
feat(deps): migrate from execa to tinyexec

ghiscoding:
## Description

replace `execa` dependency with a much lighter alternative `tinyexec` that has no external dependencies

## Motivation and Context

I was waiting for Tinyexec to get a Synchronous API before doing the move and it is now available with their latest version, I was able to make the switch with a lot of help from AI
https://github.com/tinylibs/tinyexec#synchronous

This removes 22 dependencies 
https://npmgraph.js.org/?q=execa - 22 dependencies
https://npmgraph.js.org/?q=tinyexec - 0 dependencies

Also decrease install size quite a lot
https://node-modules.dev/grid/depth#install=execa -  623.55Kb
https://node-modules.dev/grid/depth#install=tinyexec - 35Kb

## How Has This Been Tested?

migrated all unit tests from execa to tinyexec

## Types of changes

- [ ] Chore (change that has absolutely no effect on users)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #1374, opened 2026-07-22
fix(core): forward maxBuffer through nodeOptions so large git output doesn't throw ENOBUFS

### LINKED ISSUE #1373: hasTags throws ENOBUFS on repos with many tags (maxBuffer not forwarded to nodeOptions) → silent "Assuming all packages changed"
## Describe the bug

Since the migration from `execa` to `tinyexec` (#1316, v5), `maxBuffer` is no longer honored for synchronous git calls. As a result, on a repo with many git tags, `hasTags` throws internally and the error is swallowed as `ENOTAGS`, forcing the `"Assuming all packages changed"` fallback — incremental version/publish is silently broken.

Observed logs:

```
lerna-lite WARN ENOTAGS No git tags were reachable from this branch!
lerna-lite info Assuming all packages changed
```

## Root cause

In independent mode, `collectUpdates` → `hasTags` runs:

```
git tag --list '*@*'
```

On a monorepo with tens of thousands of tags this output is large (e.g. ~54,500 tags ≈ 1.77 MB in our case). `execSync` → `xSync` (tinyexec) relies on Node's `spawnSync`, and **tinyexec only forwards `maxBuffer` through `nodeOptions`**. But `_mapOptions` (in `packages/core/src/child-process.ts`) spreads `maxBuffer` at the top level of the options object:

```ts
const { cwd, env, nodeOptions, ...rest } = opts || {};
return {
  ...rest,               // <-- maxBuffer ends up here, ignored by tinyexec
  ...
  nodeOptions: { cwd, env, shell: useShell, ...nodeOptions },
};
```

So `maxBuffer` never reaches `spawnSync`, which falls back to Node's **1 MB** default → `ENOBUFS` → thrown → caught in `hasTags` as `ENOTAGS`.

Two consequences:
1. The `--max-buffer` CLI option is effectively a no-op for these sync git calls.
2. There is no large default (execa used 100 MB), so any git command emitting > 1 MB fails.

## Reproduction

```js
import { xSync } from 'tinyexec';

// top-level maxBuffer (what lerna-lite does today) -> throws
xSync('git', ['tag', '--list', '*@*'], { maxBuffer: 100 * 1024 * 1024 });
// Error: spawnSync git ENOBUFS

// inside nodeOptions -> works
xSync('git', ['tag', '--list', '*@*'], { nodeOptions: { maxBuffer: 100 * 1024 * 1024 } });
// exitCode 0
```

(Requires a repo whose `git tag --list '*@*'` output exceeds 1 MB.)

## Expected behavior

`hasTags` succeeds, incremental change detection works (`Looking for changed packages since <tag>`), and `--max-buffer` is honored.

## Fix

Destructure `maxBuffer` in `_mapOptions` and place it inside `nodeOptions`, defaulting to 100 MB (the pre-tinyexec execa default). PR incoming.

## Environment

- `@lerna-lite/*` v5.4.1 (also present on `main`)
- Bun-managed monorepo, `version: independent`, `conventionalCommits: true`

LoicEvenium:
## Description

Fixes #1373.

`tinyexec` only forwards `maxBuffer` to Node's `spawn`/`spawnSync` through `nodeOptions`. `_mapOptions` was spreading `maxBuffer` at the top level of the options object (via `...rest`), where tinyexec ignores it. Two consequences:

1. The `--max-buffer` option is a no-op for sync git calls.
2. There is no large default (execa used 100 MB), so any git command emitting more than Node's 1 MB default throws `ENOBUFS`.

The most visible symptom: in independent mode `hasTags` runs `git tag --list '*@*'`. On a repo with tens of thousands of tags that output exceeds 1 MB → `ENOBUFS` → thrown → caught in `hasTags` as `ENOTAGS` → `collectUpdates` falls back to `"Assuming all packages changed"`, silently breaking incremental version/publish.

## Motivation and Context

Regression from the execa → tinyexec migration (#1316). Before v5, execa's 100 MB `maxBuffer` default meant this never surfaced.

## How Has This Been Tested?

- Repro confirmed: `xSync('git', ['tag','--list','*@*'], { maxBuffer: 100*1024*1024 })` throws `ENOBUFS`; the same with `{ nodeOptions: { maxBuffer: ... } }` succeeds.
- Applied the equivalent change as a `bun patch` on `@lerna-lite/core@5.4.1` in a real independent monorepo (~54.5k tags): `hasTags('*@*')` now returns `true` and incremental detection is restored.

Happy to add a unit test around `_mapOptions` / `nodeOptions` propagation if you'd like — let me know your preferred shape (the fn isn't currently exported).

## Types of changes

## Changes

`packages/core/src/child-process.ts` — `_mapOptions`: destructure `maxBuffer` and place it inside `nodeOptions`, defaulting to `100 * 1024 * 1024` (matches the previous execa default).

### DIFF
--- a/packages/core/src/child-process.ts
+++ b/packages/core/src/child-process.ts
@@ -274,7 +274,7 @@ function _createEnhancedError(result: any, command: string, args: string[] = [])
 
 /** Maps Lerna/TinyExec options to tinyexec Options format */
 function _mapOptions(command: string, opts?: TinyExecOptions): Options {
-  const { cwd, env, nodeOptions, ...rest } = opts || {};
+  const { cwd, env, nodeOptions, maxBuffer, ...rest } = opts || {};
 
   // Only use shell for the 'exit' command (used in status tests)
   // Using shell: true for 'git commit' causes arguments with spaces to break.
@@ -291,6 +291,14 @@ function _mapOptions(command: string, opts?: TinyExecOptions): Options {
       cwd,
       env,
       shell: useShell,
+      // tinyexec only forwards maxBuffer to Node's spawn/spawnSync through nodeOptions,
+      // so it must live here rather than at the top level (where the `--max-buffer`
+      // option currently lands, silently ignored). Without a large default, sync git
+      // calls that emit big output overflow Node's 1MB default and throw ENOBUFS —
+      // e.g. `git tag --list '*@*'` in independent mode on repos with tens of thousands
+      // of tags, which hasTags swallows as ENOTAGS -> "Assuming all packages changed".
+      // Default to 100MB to match the pre-tinyexec (execa) behavior.
+      maxBuffer: maxBuffer ?? 100 * 1024 * 1024,
       ...nodeOptions,
     },
   } as any;


ghiscoding:
Thanks a lot, I'll push a release later today or tomorrow. Merci ⭐ 

Side note, I asked copilot to add unit tests which it did in PR #1376 via `execSync()` which uses `_mapOptions`
Encore une fois merci 😄 ⭐

